PR #1, opened 2017-05-26
Initial creation of files via ras-swagger-codegen

jonmorphic:


LJBabbage:
The code generation tool needs a number of tweaks to get rid of unnecessary imports and redundant code, however as a starting place this is ok to merge

LJBabbage (approved):

PR #2, opened 2017-05-31
Add models & break cyclic dependency between model and ONSEnvironment

jonmorphic:


README.md
```diff
@@ -24,6 +26,11 @@ Your Swagger definition lives here:
 http://localhost:8080/party-api/1.0.0/swagger.json
 ```
 
+## Database
+
+The database will automatically be created when starting the application. It is also possible to drop/create the
+database via create_db.py. Run `python create_db.py` in the project root to see further guidance.
```
LJBabbage: python create_db.py needs changing now

swagger_server/database.py
```diff
@@ -0,0 +1,71 @@
...
+
+def drop(config, engine, logger):
+    db_connection = config.get('db_connection')
+    db_schema = config.get('db_schema')
+
+    from swagger_server.models_local.base import Base
+    # fix-up the postgres schema:
+    Base.metadata.schema = db_schema if db_connection.startswith('postgres') else None
+
+    from swagger_server.models_local import _model
```
LJBabbage, line 15: As discussed one of these needs to change, ref line 33

LJBabbage (changes requested):
LGTM, bar a couple of minor changes

swagger_server/database.py
jonmorphic: Fixed. This is likely to change again in a subsequent PR as ras-party-v2 drifts away from code generator.

LJBabbage (approved):

PR #5, opened 2017-06-08
Jon/migrate from v1

jonmorphic:


swagger_server/controllers_local/controller.py
```diff
@@ -0,0 +1,524 @@
...
+    """
+    adds a reporting unit of type Business
+    Adds a new Business, or updates an existing Business based on the business reference provided
+    :param party: Business to add
+    :type party: dict | bytes
+
+    :rtype: None
+    """
+
+    v = validate.Validator(validate.Exists('id', 'reference'), validate.IsUuid('id'))
```
LJBabbage: Like the pattern, just feel this line is difficult to read and doing too much, maybe split it up into its parts e.g. not have a function inside a function. Also if we could change one of these names it would help my brain, as effectively we are doing validate.Validator(..).validate

swagger_server/controllers_local/controller.py
```diff
@@ -0,0 +1,524 @@
...
+    if not business:
+        db_address = Address(**address)
+        business = Business(ru_ref, db_party, db_address)
+        db.session.add(business)
+
+    business.attributes = party.get('attributes', {})
+
+    if associations:
+        for assoc in associations:
+            assoc_type = assoc['sampleUnitType']
```
LJBabbage: assoc_type isn't used

swagger_server/controllers_local/controller.py
```diff
@@ -0,0 +1,524 @@
...
+    v = validate.Validator(validate.Exists('id', 'reference'), validate.IsUuid('id'))
+    if not v.validate(party):
+        return make_response(jsonify(v.errors), 400)
+
+    party_uuid = party['id']
+    ru_ref = party['reference']
+    associations = party.get('associations')
+    address = party.get('address')
+
+    db_party = db.session.query(Party).filter(Party.party_uuid == party_uuid).first()
```
LJBabbage, line 53: Is this worth extracting to make it reusable?

swagger_server/controllers_local/controller.py
```diff
@@ -0,0 +1,524 @@
...
+            assoc_party = db.session.query(Party).filter(Party.party_uuid == assoc_id).first()
+            # TODO: deal with missing party
+            # TODO: assumes a BI association, implied by the table BusinessRespondent, but can there be others?
+            business_respondent = BusinessRespondent()
+            business_respondent.respondent = assoc_party.respondent
+            business_respondent.business = business
+            # TODO: check the association doesn't already exist?
+            db.session.add(business_respondent)
+
+    db.session.commit()
```
LJBabbage, line 84: One for the future but is it worth thing about some kind of  session management. I implemented the context_manager one should on this page http://docs.sqlalchemy.org/en/latest/orm/session_basics.html

swagger_server/test_local/test_controller.py
```diff
@@ -0,0 +1,329 @@
+# coding: utf-8
```
LJBabbage, line 1: Good work with the tests

LJBabbage (commented):
Away the rest of this week so not to block merge, left as comments, only a few points good work

swagger_server/controllers_local/controller.py
jonmorphic: I've imported the classes from the validate module, which strips out a level of namespacing.

swagger_server/controllers_local/controller.py
jonmorphic: We're using scoped_session.

oddjobz:
Ok, I think a huge amount of this is autogen boilerplate so as this is "new" I'm happy in general to work on the basis that any problems will be picked up quickly once we start hitting endpoints.

oddjobz (approved):
Would be useful to get a list of implemented endpoints, if we could comment out the unused part of the swagger spec it might make testing easier ...

PR #6, opened 2017-06-14
Jon/update post response

jonmorphic:
Changed functionality to:
* Create a new party id on POST new party, when party id is not supplied.
* Return the created resource in the response body when a new party is created.


oddjobz (approved):
..

PR #8, opened 2017-06-19
Jon/api gw

jonmorphic:
* Integrate with API gateway.
* Correctly end database sessions.
* Rudimentary catch-all error handling.
* Changes to simplify API spec to just the implemented endpoints.


swagger_server/test_local/test_util.py
```diff
@@ -0,0 +1,23 @@
...
+            "other": {
+                "name": "Bar"
+            }
+        }
+
+        actual = flatten_keys(d)
+
+        expected = ['name', 'other', 'other.name']
+
+        #self.assertEqual(actual, expected)
```
LJBabbage: Worth removing?

LJBabbage (approved):
I little bit of documentation would be good, specially in controllers and utils, but the rest is good

swagger_server/test_local/test_util.py
jonmorphic: Good spot, assertCountEqual should be used here.

PR #9, opened 2017-06-21
Updates for running with 0.3 API GW

oddjobz:
This introduced dependencies based on the new common code but in particular on the API GW. This version needs to run against an 0.3+ version of the GW. I've also removed the old security code that we're no longer using (from the swagger file) which makes the startup far less noisy ..)

### DIFF
--- a/config.ini
+++ b/config.ini
@@ -46,9 +46,13 @@ db_host = localhost
 ;
 ;   db_service needs to point at the name of the cloud foundry database service
 ;
-api_gateway = localhost
+api_protocol = http
+api_host = localhost
 api_port = 8080
-protocol = http
+;
+flask_protocol = http
+flask_host = localhost
+;
 db_service = ras-party-db
 ;
 ;   these settings are for development only. debug will enable the debug flag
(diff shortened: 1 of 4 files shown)


config.ini
```diff
@@ -92,6 +96,10 @@ db_connection = ${db_driver}://${db_user}:${db_pass}@${db_host}:${db_port}/${db_
 db_schema = ras
 
 [demo]
+pi_protocol = https
```
jonmorphic: Typo here?

jonmorphic (changes requested):
Looks straightforward, just a possible typo in demo section.

jonmorphic:
A further comment on this, but one to deal with out-of-band of this PR. How can I tell which is the latest stable version of ras-common? This branch pulls from a feature branch version, is that what you intended?

oddjobz:
Ok, fixed typo, and have changed the requirements.txt .. branching obviously subject to some dicussion but for now it's tagged to the 'release' branch.

jonmorphic (approved):

jonmorphic:
I did approve these changes on the basis the tests pass for me. However, I'm unsure where the syntax has come from for the ons-ras-common requirement, as it looks incorrect on closer inspection - it's actually going to pull from the head of the release/0.3 branch. I've gone down the requirements-from-git path before and it ended up a mess. I suspect tagging the version is the best we can do right now until a PyPI host can be setup.

Edit: meant to point out the Travis test failures. That doesn't happen for me locally.

PR #11, opened 2017-06-22
Ensure /parties by ref returns party representation

jonmorphic:
Return the generic party representation from the GET /parties/type/{}/ref/{} endpoint.


swagger_server/test_local/config.ini
```diff
@@ -24,4 +24,4 @@ db_schema = ras
 authentication = false
 ;db_connection = sqlite:///${db_name}
 db_connection = sqlite:///:memory:
```
oddjobz, line 26: This makes sense for small db's, saves all those latent SQLite files .. :)

oddjobz (approved):
Looks good to me .. will try to get a new copy on CF later ...

PR #12, opened 2017-06-22
Jon/update parties

jonmorphic:
Allow updates to existing business parties.


run.py
```diff
@@ -0,0 +1,4 @@
+from ons_ras_common import ons_env
```
oddjobz, line 1: Probably worth having a discussion at some point re; how micro-services are launched, doesn't worry me one way or another but it would be good to have a standard approach. I think the current mechanism was borne of Lewis's standards and paths required for the code generator, rather than a specific must-have ...

oddjobz (approved):
Yup, looks good, again will push to demo on CF later

jonmorphic:
Don’t think I intended to check-in that file! It’s useful though for running the entire app under debugger control.

PR #13, opened 2017-06-26
Jon/fix party notfound

jonmorphic:
Correctly deal with GETting a non-existent party.


oddjobz (approved):
Looks ok, one caveat; doesn't seem to pass tests on my workstation, albeit it passes elsewhere. Something to revisit and look at later ...

PR #17, opened 2017-06-27
Jon/allow business updates

jonmorphic:
Enable update of business party on either businessRef or uuid


pricem14pc (commented):
@jonmorphic I'm not familiar with the SQLAlchemy merge but am I right in thinking this change means we update the same UUID record with the latest attributes that have been reposted. I've asked Simon to test this and let us know.

jonmorphic:
@pricem14pc, merge basically does “create or update” based off the table’s primary key. In this case, the party UUID is the PK. If the given PK exists, it will do the update, “merging” in any supplied columns with the existing data. Note that the “attributes” of a business party are currently held in a single JSON column, so I think those would be overwritten wholesale. If the PK doesn’t exist, then a new row is inserted.

I’m unclear on if this is what’s actually wanted.

oddjobz:
You seem to be removing run.sh and test.sh which were the pre-agreed mechanism for running and testing new applications locally. I've no objection to taking a better approach, but could we add something to the README to indicate what they're being replaced by?

jonmorphic:
The README file states "To run the tests with tox, install tox (see link) then simply run the command tox in the root of the project". I can add further detail if this is unclear.

I'm unaware of prior discussions on this, but would strongly recommend tox over shell scripts, both as a better solution and as a well-known Python tool.

oddjobz (approved):
Looks good to go, now passing tests on my machine following an upgrade to Python3.6 .. :)

PR #20, opened 2017-06-28
Quoting keys and changing businessAssociations to associations

pricem14pc:


pricem14pc:
Hi Jon.. as per the request by RM.

jonmorphic (approved):

PR #22, opened 2017-07-03
Integration test

oddjobz:
Minor environmental changes for deployment onto "test" and "demo". Note following the merge the tracking for ras-common should probably reset to master (and a fixed version?)


jonmorphic:
Couple of comments:

- Would be good to move away from having environment config in service repos. I have a proposed solution which we can discuss out-of-band of this PR.
- Unclear on the mechanism of resetting the ras-common branch following merge, as that requires a subsequent PR to make the change. Presumably we can just do the ras-common PR first?

oddjobz:
I would agree that too much configuration in public repo's is bad. However, separating all default configuration from the code is also not ideal. Not really sure what the answer is given "make it private" doesn't fly. 

This was always the problem with the release/0.3 branch from which all this stems .. but I think you're right, doing ras-common first is probably the best approach.

jonmorphic:
Not sure I fully understand your reply, although that needn't hold up this PR. I'm primarily referring to environmental config (as opposed to app config), which has a different lifecycle to services. This is a really good rule-of-thumb for partitioning things into repositories. For example, we wouldn't want to end up in a situation where to add a new environment we need to update, release and deploy multiple services.

I've merged the ras-common PR. Once you're happy, can you release that repo and reference the tag here?

oddjobz:
Mmm, good point .. we could certainly move the environment specific components to a different config.ini just leaving travis and development for testing and local.

(release which repo?)

jonmorphic:
Was referring to releasing the ras-common repo which I've just merged. For expediency, I've just release it. So now the github reference for requirements.txt is git+https://github.com/ONSdigital/ras-common.git@0.1.114

oddjobz:
Ok have re-tagged requirements.txt to master ...

jonmorphic (approved):

PR #27, opened 2017-07-13
Feature/us007 7b get case context

jonmorphic:
Includes three strands of work:

- Introduction of the OAuth2 registration process for creating a new respondent/user.
- Integration with RM services in order to obtain information to create a business-respondent association and an enrolment.
- Integration of ras-common-utils, which is a stripped-down equivalent of ras-common, allowing ras-party to adopt a conventional Flask app structure. Configuration values are moved into config.yaml.

Note the changes are ongoing beyond this PR, so this represents a checkpoint. Subsequent PRs will follow-up with the extra tasks to post a case event, generate a verification email and expose a new endpoint to retrieve party associations.


LJBabbage (approved):
Agreed to merge and tidy up in 2 separate PR's to keep story moving and stop massive commit history growing even further. Points to address 
1. Test coverage
2. Logging 
3. Noise in code
4. Loose ends and TODO's

PR #31, opened 2017-07-19
Feature/us007 case event

jonmorphic:
- refactor to simplify the project structure.
- refactor to simplify the error handling and reduce the amount of inline error handling / logging
- generate an email verification token (but don't call gov.uk notify yet)
- introduce email verification endpoint
- post a case event

Note: Code from Nick's branch to carry out user activation is also integrated into this PR. See https://github.com/ONSdigital/ras-party/pull/33


LJBabbage (approved):
After discussion with team lead this has been agreed to merge. I have only been able to do a very quick overview and not test the PR fully.  As such I can't confirm this works as expected

PR #32, opened 2017-07-20
Add uk gov notify and tests

LJBabbage:
### What is the context of this PR?
To add in the functionality to use gov.uk notify. This has been isolated away from the code and will be plugged in later. 

### How to review 
At present a live test is not possible as there is no key management process, meaning the real keys aren't represent on the repo. So this review is code structure more then anything else

### DIFF
--- a/swagger_server/controllers/gov_uk_notify.py
+++ b/swagger_server/controllers/gov_uk_notify.py
@@ -1,0 +1,37 @@
+from flask import current_app
+from notifications_python_client.notifications import NotificationsAPIClient
+from structlog import get_logger
+from swagger_server.controllers.ras_error import RasNotifyError
+
+log = get_logger()
+
+
+class GovUKNotify:
+    """ Gov uk notify class"""
+
+    def __init__(self):
+        notify_service = current_app.config.dependency['gov-uk-notify-service']
+        notify_keys = 'key-name-{}-{}'.format(notify_service['gov_notify_service_id'],
+                                              notify_service['gov_notify_api_key'])
+        self.notifications_client = NotificationsAPIClient(notify_keys)
+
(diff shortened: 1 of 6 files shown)


swagger_server/controllers/gov_uk_notify.py
```diff
@@ -0,0 +1,37 @@
...
+         """
+        try:
+            response = self.notifications_client.send_email_notification(
+                email_address=email,
+                template_id=template_id,
+                personalisation=personalisation,
+                reference=reference)
+            log.info('Message sent to gov.uk notify with notification id {}'.format(response.id))
+        except Exception as e:
+            log.error('Gov uk notify can not send the message' + str(e))
```
jonmorphic, line 35: I've tied the error logging in with the exception handling, so for example if you did:
`raise RasNotifyError("Gov uk notify cannot send the message", status_code=400)`
The text should get logged with error level. Your call on how useful this is.

swagger_server/test/test_gov_uk_notify.py
```diff
@@ -0,0 +1,38 @@
...
+            # Then a message is created
+            self.assertEquals(response, 201)
+
+    def test_notify_exception(self):
+
+        # Given a mocked gov.uk notify and exception
+        notify_patch = Mock()
+        notify_patch.send_email_notification = Mock(side_effect=Exception)
+
+        with patch('swagger_server.controllers.gov_uk_notify.NotificationsAPIClient', return_value=notify_patch):
```
jonmorphic, line 33: Pedantic point: May be worth considering the patch decorator, should be able to reduce the mocking by a couple of lines.

jonmorphic (approved):
Looks fine to me. Couple of minor observations, not necessary to fix here, thus approved.

PR #46, opened 2017-08-13
Consistent email verification log messages and returning 200 if user …

pricem14pc:
…already verified (instead of 409 error)


jonmorphic:
One comment: flake8 issues introduced, these can be seen by running tox from root of project (see README.md). Have pushed a fix for these.

jonmorphic (approved):

PR #49, opened 2017-08-15
Add end point for verification email and refactor existing code

LJBabbage:
### What is the context of this PR?
To add an end point for resending of verification emails and refactor the code so the functionality isn't locked as it was before. As the story is not fully formed I have made some assumption
1. The end point is called /resend-verification-email/<party_uuid>
2. It can return 3 responses - email resent, respondent already verified, respondent not found
3. It uses party_uuid


ras_party/controllers/controller.py
```diff
@@ -33,6 +34,10 @@
     REQUESTS_GET_TIMEOUT = 20
```
jonmorphic, line 34: I'm not sure this code is part of the PR, but I've never seen it before, and am wondering why we need to catch a generic Exception here? AFAICT, there's only a ValueError to worry about, which we can just allow to propagate. Also, we really need to avoid having direct access into the environment / file system from arbitrary locations in the code. There's an existing mechanism to inject config from the top of the process.

ras_party/controllers/controller.py
```diff
@@ -531,8 +526,86 @@ def post_case_event(case_id, party_id, category="Default category message", desc
...
+    """
+    Check and resend an email verification email
+    :param party_uuid: the party uuid
+    :return: make_response
+    """
+    log.debug('attempting to resend verification_email with party_uuid {}'.format(party_uuid))
+
+    respondent = _query_respondent_by_party_uuid(party_uuid)
+
+    if not respondent:
```
jonmorphic, line 551: A minor point, may be worth considering raising an exception for the unhappy path flow which will log the message and return a formatted error response.

test/test_controller.py
```diff
@@ -200,7 +201,7 @@ def test_post_respondent_creates_the_enrolment(self, _):
         self.assertEqual(str(enrolment.business_respondent.business.party_uuid),
                          '3b136c4b-7a14-4904-9e01-13364dd7b972')
 
-    @patch('ras_party.controllers.controller.notify')
+    @patch('ras_party.controllers.controller._send_message_to_gov_uk_notify')
```
jonmorphic, line 204: Maybe one for our tech debt list, would be good to avoid patching something within the unit-under-test, and to patch at the level of the GovUKNotify class instead (I know that it was that way before :)). Uncomfortable about seeing a patch of a "protected" name.

jonmorphic (commented):
Generally happy with the PR, subject to the stated assumptions. Is there a supporting story for this which covers the end-to-end feature?

jonmorphic (approved):

PR #50, opened 2017-08-16
Integrate 0.0.5 of ras-common-utils

jonmorphic:
This integrates v0.0.5 of ras-common-utils, which provides fixes for correctly looking-up config in the CloudFoundry environments. Also integrates the RasError class from ras-common-utils and removes redundant code.


ras_party/controllers/controller.py
```diff
@@ -3,13 +3,13 @@
 import os
 from flask import make_response, jsonify, current_app
 from itsdangerous import URLSafeTimedSerializer, BadSignature, BadData, SignatureExpired
+from ras_common_utils.ras_error.ras_error import RasError, RasNotifyError
```
nherriot, line 6: Just a pickie point. Keeping all imports together for that subgroup. Could move this down to line 10.
Also you have moved errors into ras_common - why not leave this for the time being then once buy in from the team is received it can be used by other microservices?

ras_party/controllers/error_decorator.py
```diff
@@ -2,9 +2,9 @@
 
 import structlog
 from flask import current_app
+from ras_common_utils.ras_error.ras_error import RasError
```
nherriot, line 5: Same as other comment - a pickie point i know and only change if you have time and can be arsed.

nherriot (approved):
This pull request on the trello board is for how detection is done, but I can only see re factoring of ras_common. Was this what you meant?

ras_party/controllers/controller.py
jonmorphic: Thanks Nick. I normally let PyCharm do the imports.

jonmorphic:
Re the last comment, this PR pulls in 0.0.5 of ras-common-utils, which contains the fix for CloudFoundry config.

PR #53, opened 2017-09-04
Isu 17 incorrect url in verification email

nherriot:
This allows the email URL to be added via the config.yml file.
Also has a test case to check it exists.


ras_party/controllers/controller.py
```diff
@@ -569,6 +570,7 @@ def _send_email_verification(party_uuid, email):
     """
 
     verification_url = _create_verification_url(email)
+    # verification_url = "http://this-is-a-test"
```
LJBabbage: I can't see any value in keeping this comment

config/config.yaml
```diff
@@ -30,12 +30,13 @@ service:
...
     PARTY_SCHEMA: ras_party/schemas/party_schema.json
+    PUBLIC_EMAIL_VERIFICATION_URL: http://ons.survey.gov/bres                   # Do not add trailing slash e.g. http://x.y/parent/child/ BAD http://x.y/parent/child GOOD
 
 dependencies:
     ras-party-db:
         schema: partysvc
-        #uri: "sqlite:///ras-party"
-        uri: "postgresql://postgres:postgres@localhost:5432/postgres"
+        uri: "sqlite:///ras-party"
+        #uri: "postgresql://postgres:postgres@localhost:5432/postgres"
```
LJBabbage: Not sure why this comment is here

config/config.yaml
```diff
@@ -30,12 +30,13 @@ service:
     SECRET_KEY: aardvark
     EMAIL_TOKEN_SALT: aardvark
     PARTY_SCHEMA: ras_party/schemas/party_schema.json
+    PUBLIC_EMAIL_VERIFICATION_URL: http://ons.survey.gov/bres                   # Do not add trailing slash e.g. http://x.y/parent/child/ BAD http://x.y/parent/child GOOD
```
LJBabbage: We could write this comment for any url in our system, lets not start doing that

test/test_controller.py
```diff
@@ -230,16 +231,41 @@ def test_email_verification_activates_a_respondent(self, _, mock_notify):
...
         # When the email is verified
         frontstage_url = mock_notify.call_args[0][2]
         token = frontstage_url.split('/')[-1]
         self.put_email_verification(token, 200)
-
         # Then the respondent state is ACTIVE
         db_respondent = respondents()[0]
         self.assertEqual(db_respondent.status, RespondentStatus.ACTIVE)
 
+    # This tests that the email verification URL is dynamically taken from a parameter in the configuration called
```
LJBabbage: Good naming of the test and the Given, When, Then structure should mean we don't need any further comments

test/test_controller.py
```diff
@@ -230,16 +231,41 @@ def test_email_verification_activates_a_respondent(self, _, mock_notify):
...
+        # Send POST to /respondents endpoint which adds a user to the DB (which generates an email verification)
+        self.post_to_respondents(mock_respondent, 200)
+        # And the respondent state is CREATED
+        db_respondent = respondents()[0]
+        self.assertEqual(db_respondent.status, RespondentStatus.CREATED)
+        # Load the public email verfication URL from our test config file. make sure this is the URL used when sending
+        # out the verification email to the notify service
+        config_data = yaml.load(test_config)
+        test_url = config_data['service']['PUBLIC_EMAIL_VERIFICATION_URL']
+        # print("***** config values are: {}".format(test_url))
```
LJBabbage: This can be removed

test/test_controller.py
```diff
@@ -230,16 +231,41 @@ def test_email_verification_activates_a_respondent(self, _, mock_notify):
...
+
+        # When the email is verified get the email URL from the argument list in the '_send_message_to_gov_uk_notify'
+        # method
+        frontstage_url = mock_notify.call_args[0][2]
+        self.assertIn(test_url, frontstage_url)
```
LJBabbage: #Then ...

config/config.yaml
LJBabbage: Not a fan of using defaults like this. If Jenkins is configured wrong the user is going to start getting them without us knowing, can't be good

LJBabbage (changes requested):

ras_party/controllers/controller.py
nherriot: removed

config/config.yaml
nherriot: changed to ardvark

config/config.yaml
nherriot: it's a way to get the system to run locally. reversed them, so back to normal

test/test_controller.py
nherriot: done.

LJBabbage (approved):

config/config.yaml
```diff
@@ -30,6 +30,7 @@ service:
     SECRET_KEY: aardvark
     EMAIL_TOKEN_SALT: aardvark
     PARTY_SCHEMA: ras_party/schemas/party_schema.json
+    PUBLIC_EMAIL_VERIFICATION_URL: aardvark                   # Do not add trailing slash e.g. http://x.y/parent/child/ BAD http://x.y/parent/child GOOD
```
jonmorphic: It looks like the frontstage dependency is redundant - can we remove it from the config file.

jonmorphic (changes requested):

config/config.yaml
nherriot: I'll add as a TODO and put in release notes as i need to merge this for 5.  don't want to break anything.

jonmorphic (approved):

PR #55, opened 2017-09-12
Feature/us167 internal change respondents email address

andrewmil:
- Added endpoint  `PUT /respondents/email` which can be used to change user's email address(username)
- added function `update_oauth_user` which can update any oauth value for users.
- Added a couple of tests around this functionality

### DIFF
--- a/ras_party/views/party_view.py
+++ b/ras_party/views/party_view.py
@@ -79,6 +79,19 @@ def get_respondent_by_email(email):
     return controller.get_respondent_by_email(email)
 
 
+@party_view.route('/respondents/email', methods=['PUT'])
+@auth.login_required
+@log_route
+def change_respondent_email():
+    request_data = request.json
+    email_address = request_data.get('email_address')
+    new_email_address = request_data.get('new_email_address')
+    if not email_address or not new_email_address:
+        return make_response(jsonify({'errors': 'No email provided'}), 400)
+
+    return controller.change_respondent_email(email_address, new_email_address)
+
+
 @party_view.route('/respondents', methods=['POST'])
 @auth.login_required
 @log_route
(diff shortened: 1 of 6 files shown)


ras_party/views/party_view.py
```diff
@@ -79,6 +79,19 @@ def get_respondent_by_email(email):
...
+@auth.login_required
+@log_route
+def change_respondent_email():
+    request_data = request.json
+    email_address = request_data.get('email_address')
+    new_email_address = request_data.get('new_email_address')
+    if not email_address or not new_email_address:
+        return make_response(jsonify({'errors': 'No email provided'}), 400)
+
+    return controller.change_respondent_email(email_address, new_email_address)
```
tej99, line 92: Can change-respondent_email handle a None for email_address? Also session isn't passed here, maybe set a default for session.

andrewmil: If the email_address is `None` then the 400 error on line 90 will trigger
And the session actually get's passed to `controller.change_respondent_email` by its decorator, `with_db_session`

andrewmil: Actually I think you mean how the function itself would deal with a None for email_address, I'm not 100% sure whether the database query on line 642(!) of controller.py would fail or return None

jonmorphic: I'd be inclined to pass the entire payload through to the controller function and add the defensive code there. For example, see respondents_post for validating required fields.

jonmorphic (approved):
Generally looks good to me, UT coverage is 88%.

I'm about to refactor a lot of the controller logic as part of US012, happy to merge this PR first and deal with conflicts on my branch.

jonmorphic:
Just noticed the codacy issue - safe to ignore.

PR #57, opened 2017-09-15
Fix the construction of public website urls

jonmorphic:


LJBabbage (changes requested):
All fine, just would like the tests split up and if we could as team agree to do the Given, When, Then pattern it would be awesome

test/support/test_public_website.py
```diff
@@ -23,39 +23,138 @@ def __getitem__(self, item):
 
 class TestPublicWebsite(TestCase):
 
-    def test_reset_password_url(self):
-
-        config = MockConfig()
+    def test_reset_password_url_excludes_port_80_when_scheme_is_http(self):
```
andrewmil: This is really 3 tests in 1 splitting them up would probably make it more readable

test/support/test_public_website.py
```diff
@@ -23,39 +23,138 @@ def __getitem__(self, item):
...
+
+        # When the reset password url is constructed
+        actual_url = unit.reset_password_url('test@email.com')
+
+        # Then it does not contain the port number
+        expected_url_substring = "http://mockhost/passwords/reset-password/"
+        self.assertIn(expected_url_substring, actual_url)
+
+    def test_reset_password_url_excludes_port_443_when_scheme_is_https(self):
+        # Given config contains SCHEME 'http' and PORT 443
```
andrewmil: The comments here say `http` instead of `https`

andrewmil (approved):
Looks like things are working and the testing is comprehensive.

Small comment inline about the structure of the tests but functionally all looks good

test/support/test_public_website.py
jonmorphic: One of the drawbacks of non-executable Gherkin-like syntax :)

LJBabbage (approved):

PR #58, opened 2017-09-15
removing active respondent check for resend email

andrewmil:
Removing a restriction to allow users to have their verification email sent if their status is set to `ACTIVE`

Removed a test related to this

### DIFF
--- a/ras_party/controllers/party_controller.py
+++ b/ras_party/controllers/party_controller.py
@@ -22,7 +22,6 @@ from ras_party.support.verification import decode_email_token
 log = get_logger()
 
 NO_RESPONDENT_FOR_PARTY_ID = 'There is no respondent with that party ID '
-EMAIL_ALREADY_VERIFIED = 'The Respondent for that party ID is already verified'
 EMAIL_VERIFICATION_SENT = 'A new verification email has been sent'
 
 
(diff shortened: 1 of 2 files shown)


ras_party/controllers/party_controller.py
```diff
@@ -623,17 +622,12 @@ def resend_verification_email(party_uuid):
     :param party_uuid: the party uuid
     :return: make_response
     """
-    log.debug('attempting to resend verification_email with party_uuid {}'.format(party_uuid))
+    log.debug('Attempting to resend verification_email with party_uuid {}'.format(party_uuid))
```
rhiGriff, line 625: Does this need to be updated to conform with the logging standards?

andrewmil: It does but all the other logging in party service needs to be updated to meet those standards as well I think. Maybe we should add to the technical debt?

rhiGriff (approved):
Happy now, I think the changes to logging should be raised as technical debt

jonmorphic:
What logging standards?

andrewmil:
https://digitaleq.atlassian.net/wiki/spaces/SDC/pages/69632229/Logging

I don't think they've been well communicated across the service.
Pretty much it boils down to having everything in key-value pairs so splunk can use it.

Since we're using structlog we could update that log line to be;
`log.debug('Attempting to resend verification_email', party_uuid=party_uuid)`

nherriot (commented):
Looks good.

PR #59, opened 2017-09-21
Dockerise party service

jwgwalton:
Contains script to spool up a local dev environment of a postgres container with the party-service and populate with test data. Working towards fully dockerising each component to have a working ras-compose

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM python:3.4
+WORKDIR /app
+COPY . /app
+EXPOSE 8081
+RUN pip3 install -r requirements.txt
+ENTRYPOINT ["python3"]
+CMD ["run.py"]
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM python:3.4
```
jonmorphic: Why not Python 3.6?

jonmorphic:
One potential issue, there's no guarantee the DB is up and responding before party service tries to initialise the model. However, it's unlikely this can be resolved within the scope of this PR, as it ties into the wider issue of how a service acquires and maintains its dependencies.

liamtoozer (commented):
Functionally all good; spins up container fine for me.

jonmorphic (approved):

PR #60, opened 2017-09-25
Email address exposed in some urls and logs

andrewmil:
`get_respondent_by_email` endpoint now receives an encrypted email token rather than a standard email string.

This is then decoded in the same way as the other `token` endpoints in party services.

Factored out a `get_email_from_token` function which was used in several places

This change works in conjunction with https://github.com/ONSdigital/ras-frontstage/pull/211

https://trello.com/c/dUefc87O/667-email-address-exposed-in-some-urls-and-logs

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -39,7 +39,7 @@ service:
 dependencies:
     ras-party-db:
         schema: partysvc
-        uri: "sqlite:///ras-party"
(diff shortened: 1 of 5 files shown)


LJBabbage (changes requested):
Further issues need discussion before this can be merged

PR #63, opened 2017-10-18
TBL098 (P3) Verbose Business Party

tej99:
Link to trello: https://trello.com/c/0goz2j83/710-tbl098-p3-verbose-business-party

Changes made:
-  Default GET requests to /businesses/ref/<ref> and /businesses/id/<id> endpoints now return a summary of business details
- Added 'verbose' param to the endpoints. ?verbose=true returns all business details
- Changed JSON structure for GET /parties/type/<sampleUnitType> endpoints to return 'name' on the outer layer instead within the 'attributes' key (It will still be there too, for the time being)

### DIFF
--- a/API.md
+++ b/API.md
@@ -56,11 +56,11 @@ This page documents the Party service API endpoints. All endpoints return an `HT
         "cell_no": 1,     
         "formtype": "0001",
         "currency": "H",
-        "name": "Bolts and Ratchets Ltd"
     },
     "id": "b3ba864b-7cbc-4f44-84fe-88dc018a1a4c",
     "sampleUnitRef": "50012345678",
-    "sampleUnitType": "B"
+    "sampleUnitType": "B",
+    "name": "Bolts and Ratchets Ltd"
 }
 ```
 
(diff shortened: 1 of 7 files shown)


ras_party/models/models.py
```diff
@@ -83,21 +83,35 @@ def _get_respondents_associations(respondents):
...
 
-    def to_business_dict(self):
-        d = {
-            'id': self.party_uuid,
-            'sampleUnitRef': self.business_ref,
-            'sampleUnitType': self.UNIT_TYPE,
-            'associations': self._get_respondents_associations(self.respondents),
-        }
-        return dict(d, **self.attributes)
+    def to_business_dict(self, summary=False):
```
benjefferies: Is it worth splitting this logic out into `to_business_dict` and `to_business_summary_dict` to create more cohesive easier to test methods?

ras_party/controllers/party_controller.py
```diff
@@ -27,42 +27,46 @@
...
         raise RasError("Business with reference '{}' does not exist.".format(ref), status_code=404)
 
-    return business.to_business_dict()
+    return business.to_business_dict(summary=not verbose)
 
 
 @translate_exceptions
 @with_db_session
-def get_business_by_id(id, session):
+def get_business_by_id(party_uuid, session, verbose=False):
```
benjefferies, line 53: Is it worth splitting this logic out into `get_business_by_id` and `get_business_summary__by_id ` to create more cohesive easier to test methods?

API.md
```diff
@@ -56,11 +56,11 @@ This page documents the Party service API endpoints. All endpoints return an `HT
         "cell_no": 1,     
         "formtype": "0001",
         "currency": "H",
-        "name": "Bolts and Ratchets Ltd"
```
benjefferies, line 59: Does this make it a breaking change by removing the name from the attribute?

benjefferies (changes requested):

ras_party/models/models.py
```diff
@@ -83,21 +83,35 @@ def _get_respondents_associations(respondents):
...
+
+        if summary:
```
andrewmil: You can shorten this bit of logic, define `d` outside the if statement with the common fields and have 
`if summary -> d[associations] = ...`

ras_party/controllers/party_controller.py
```diff
@@ -27,42 +27,46 @@
...
+    :param verbose: Verbosity of business details
+
     :rtype: Business
     """
     business = session.query(Business).filter(Business.business_ref == ref).first()
     if not business:
         raise RasError("Business with reference '{}' does not exist.".format(ref), status_code=404)
 
-    return business.to_business_dict()
+    return business.to_business_dict(summary=not verbose)
```
andrewmil: would it be simpler to flip the logic for summary and just call it verbose too? Then we don't have to invert the logic half way through

ras_party/models/models.py
andrewmil: related to the above you could call this verbose too and have it the other way around

andrewmil (changes requested):
Left a couple of small inline comments.
Also noted that there isn't much being logged for this journey I don't think, should we add more?

API.md
tej99: In the short term that attribute will be duplicated in both places, but eventually 'name' will be removed from 'attributes' as part of the bigger party service refactoring. This has been raised to teams that use this endpoint.

benjefferies (approved):

andrewmil (approved):

PR #65, opened 2017-10-24
[Defect 71] Change all respondents queries to be case insensitive

benjefferies:
Refactored all email queries out into separate method which does a case insensitive query. Tests are to assert this functionality is used. Initially I intended to make sure the query was case insensitive by mocking the sqlalchemy session but the comparator appears to not implement equality checks

### DIFF
--- a/ras_party/controllers/party_controller.py
+++ b/ras_party/controllers/party_controller.py
@@ -205,7 +205,7 @@ def get_respondent_by_email(email, session):
     :param email: Email of respondent to lookup
     :rtype: Respondent
     """
-    respondent = session.query(Respondent).filter(Respondent.email_address == email).first()
+    respondent = _query_respondent_by_email(email, session).first()
     if not respondent:
         raise RasError("Respondent does not exist.", status_code=404)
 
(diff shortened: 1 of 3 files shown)


pricem14pc (commented):
I'll test this locally now.. as discussed its the same alchemy function I'd used

ras_party/controllers/party_controller.py
```diff
@@ -627,6 +625,10 @@ def _query_respondent_by_party_uuid(party_uuid, session):
     return session.query(Respondent).filter(Respondent.party_uuid == party_uuid).first()
 
 
+def _query_respondent_by_email(email_address, session):
+    return session.query(Respondent).filter(func.lower(Respondent.email_address) == email_address.lower()).one()
```
tej99: Most of the queries were making use of .first() not .one() so they expect a None type to be returned if no results are found instead of raising NoResultFound.

ras_party/controllers/party_controller.py
```diff
@@ -487,7 +485,7 @@ def put_email_verification(token, session):
         raise RasError(msg, 404)
 
     try:
-        r = session.query(Respondent).filter(Respondent.email_address == email_address).one()
+        r = _query_respondent_by_email(email_address, session)
```
tej99: sql query using .first() does not raise NoResultFound, it will return a None. Will have to replace try except with 'if not r:  raise RasError...' statement.

tej99 (changes requested):

ras_party/controllers/party_controller.py
```diff
@@ -226,14 +226,14 @@ def change_respondent(payload, tran, session):
     email_address = payload['email_address']
     new_email_address = payload['new_email_address']
 
-    respondent = session.query(Respondent).filter(Respondent.email_address == email_address).one()
+    respondent = _query_respondent_by_email(email_address, session).one()
```
tej99: Needs to be a .first() else requires a try except.

benjefferies: Good spot

pricem14pc:
Successfully tested case insensitivity locally; Created a new account, able to log in with a differently cased email address successfully, unable to create new account with a differently cased email address, requested and changed password with differently cased email address.

tej99 (approved):

pricem14pc (approved):

PR #66, opened 2017-10-26
Tbl106: use rm-notify gateway

jwgwalton:
Move from directly calling the gov uk notify service to the wrapper from RM.


jwgwalton:
Just seen test coverage diff will up this but leaving open so can see the codecov report

ras_party/controllers/gov_uk_notify.py
```diff
@@ -1,45 +1,48 @@
...
+        Send message to gov.uk notify wrapper
         :param email: email address of recipient
         :param template_id: the template id on gov.uk notify to use
         :param personalisation: placeholder values in the template
-        :param reference: reference to generated if not using Notify's id
+        :param reference: reference to be generated if not using Notify's id
         :rtype: 201 if success
         """
 
         if not self.config.feature.send_email_to_gov_notify:
```
benjefferies, line 27: Do we even need this logic here? Does this not exist in `notifygatewaysvc`?

test/test_gov_uk_notify.py
```diff
@@ -1,43 +1,62 @@
...
 
 from ras_party.controllers.gov_uk_notify import GovUkNotify
 from ras_party.support.ras_error import RasNotifyError
 from test.party_client import PartyTestClient
+from test.mocks import MockRequests
+from ras_party.support.requests_wrapper import Requests
 
 
-class TestParties(PartyTestClient):
+class TestNotify(PartyTestClient):
```
benjefferies: 👍

ras_party/controllers/gov_uk_notify.py
```diff
@@ -1,45 +1,48 @@
...
-                reference=reference)
+            notification = {
+                "emailAddress": email,
+            }
+            if personalisation:
+                notification.update({"personalisation": personalisation})
+            if reference:
+                notification.update({"reference": reference})
+
+            url = urlparse.urljoin(str(self.notify_config['url']), str(template_id))
```
benjefferies: Do we need to wrap `self.notify_config['url']` with `str`?

benjefferies (changes requested):

test/test_party_controller.py
```diff
@@ -406,6 +389,77 @@ def test_resend_verification_email_party_id_not_found(self):
...
+
+        # when the verify token endpoint is hit it errors
+        self.verify_token(token, expected_status=404)
+
+    def test_verify_token_with_valid_token(self):
+        # given a valid token
+        token = self.generate_valid_token()
+
+        # the verify end point verifies the token
+        self.verify_token(token)
```
tej99: Being explicit might read better here, plus other tests set 200 explicitly

ras_party/controllers/gov_uk_notify.py
tej99: Would you not want a toggle in this service even if there is one in notifygatewaysvc?

tej99 (approved):
Looks good 👍

ras_party/controllers/gov_uk_notify.py
benjefferies: I don't think so. The only scenario I can think of is in unit tests where you can just mock the request anyway?

andrewmil: In the current logic it shouldn't be needed.
Currently we don't check the response from the notify service so if anything failed there, (500 error or for some other reason it fails to send to gov.notify) then we'd have no way of knowing, there's not even a log line for it I don't think.
If we had that check (which I think we should) this kind of toggle would help get around that when we run party locally

benjefferies: Even then we can spin up a docker instance of a mock api which returns 200 which is defined in a docker-compose or something. It's always best to avoid putting hacks in production code to support tests or code to test locally. 

It leaves risk to running these things in production like in this scenario not sending emails when it should.

It removes unnecessary environmental/config variables used to support testing/local usage

andrewmil: Yeah makes sense, I do think we should be checking the response from the notify service at some point in party though

jwgwalton: I'm not sure if this is now defunct. I didn't think it was in for testing but to stop in different cloud foundry environments. Though as you say I think the notifygatewaysvc only has the keys to hit gov notify in prod. I'll double check with @LJBabbage or @pricem14pc if this is the case.

ras_party/controllers/gov_uk_notify.py
jwgwalton: Probably not.

ras_party/controllers/gov_uk_notify.py
benjefferies: @jwgwalton might be one for the tech debt rather than here

config/config.yaml
```diff
@@ -57,8 +57,7 @@ dependencies:
         host: localhost
         port: 8080
     notify-service:
-        api_key: notify_api_key
-        service_id: sdc_service_id
+        url: "http://notifygatewaysvc-dev.apps.devtest.onsclofo.uk/emails/"
```
LJBabbage: We default to a local version on all other services, any reason why we are using dev here?

ras_party/controllers/gov_uk_notify.py
LJBabbage: Agreed, if there is logic in the other service, we no longer need it here

test/fixtures/config.py
```diff
@@ -37,8 +37,7 @@
         host: mockhost
         port: 3333
     notify-service:
-        api_key: notify_api_key
-        service_id: sdc_service_id
+        url: http://notifygatewaysvc-dev.apps.devtest.onsclofo.uk/emails/
```
LJBabbage, line 40: Same as above

LJBabbage (changes requested):
Looks good bar the odd small comment. The only thing I didn't immediately pick up is that this is an ONS service which then manages the notify process. Things like python names (gov_uk_notify.py) and classes GovUkNotify could do with a tweek

LJBabbage (approved):

ras_party/controllers/gov_uk_notify.py
jwgwalton: @benjefferies I'm leaving this toggle in as currently there isn't one available in the rm-notify gateway.

benjefferies: Sure something we can add in as an improvement later

benjefferies (approved):

PR #67, opened 2017-10-27
Tbl088 first refactor

tej99:
Break out controllers and views to make future changes in the party service, around the enrolment process, easier.


ras_party/controllers/queries.py
```diff
@@ -0,0 +1,51 @@
+
+from sqlalchemy import func
+from structlog import get_logger
+
+from ras_party.models.models import Business, Respondent
+
+log = get_logger()
+
+
+def query_business_by_party_uuid(party_uuid, session):
```
andrewmil, line 10: Is 'get' a better word to use than 'query' in this file?

ras_party/controllers/account_controller.py
```diff
@@ -0,0 +1,461 @@
...
+
+    if not v.validate(party):
+        raise RasError(v.errors, 400)
+
+    iac = request_iac(party['enrolmentCode'])
+    if not iac.get('active'):
+        raise RasError("Enrolment code is not active.", status_code=400)
+
+    existing = query_respondent_by_email(party['emailAddress'], session)
+    if existing:
```
andrewmil, line 66: could just do `if query_respondent_by_email(party['emailAddress'], session)` since you only use existing once

ras_party/controllers/account_controller.py
```diff
@@ -0,0 +1,461 @@
...
+        raise RasError("User with email address {} already exists.".format(party['emailAddress']), status_code=400)
+
+    case_context = request_case(party['enrolmentCode'])
```
andrewmil, line 69: I do kind of get using 'query' above actually if we're looking to distinguish it from the external requests we're making like we are here

ras_party/controllers/account_controller.py
```diff
@@ -0,0 +1,461 @@
...
+    collection_exercise = request_collection_exercise(collection_exercise_id)
+
+    try:
+        survey_id = collection_exercise['surveyId']
+        survey = request_survey(survey_id)
+        survey_name = survey['longName']
+    except KeyError:
+        raise RasError("There is no survey bound for this user with email address: {}".format(party['emailAddress']))
+
+    b = query_business_by_party_uuid(business_id, session)
```
andrewmil: Should be explicit with names so should use 'business' rather than 'b'

ras_party/controllers/business_controller.py
```diff
@@ -0,0 +1,89 @@
...
+    if not business:
+        raise RasError("Business with reference '{}' does not exist.".format(ref), status_code=404)
+
+    if verbose:
+        return business.to_business_dict()
+    else:
+        return business.to_business_summary_dict()
+
+
+@translate_exceptions
```
benjefferies: Could we retire this wrapper and in favour of the flask errorhandler?

ras_party/controllers/account_controller.py
```diff
@@ -0,0 +1,461 @@
...
+    translated_party = {
+        'party_uuid': party.get('id') or str(uuid.uuid4()),
+        'email_address': party['emailAddress'],
+        'first_name': party['firstName'],
+        'last_name': party['lastName'],
+        'telephone': party['telephone'],
+        'status': RespondentStatus.CREATED
+    }
+
+    try:
```
andrewmil, line 97: Should we separate the process of adding the respondent to the database into a separate function? We have all the queries to the database separate already

ras_party/controllers/account_controller.py
```diff
@@ -0,0 +1,461 @@
...
+                  survey_name=survey_name,
+                  status=EnrolmentStatus.PENDING)
+        session.add(r)
+        session.add(pending_enrolment)
+
+        # Notify the case service of this account being created
+        post_case_event(case_id, r.party_uuid, "RESPONDENT_ACCOUNT_CREATED", "New respondent account created")
+
+        _send_email_verification(r.party_uuid, party['emailAddress'])
+    except (orm.exc.ObjectDeletedError, orm.exc.FlushError, orm.exc.StaleDataError, orm.exc.DetachedInstanceError):
```
andrewmil: I think all these exceptions inherit from `SQLAlchemyError` so we could just catch that?

ras_party/controllers/account_controller.py
```diff
@@ -0,0 +1,461 @@
...
+EMAIL_VERIFICATION_SENT = 'A new verification email has been sent'
+
+
+@translate_exceptions
+@transactional
+@with_db_session
+def post_respondent(party, tran, session):
+    """
+    This function is quite complicated, as it performs a number of steps to complete a new respondent enrolment:
+    1. Validate the enrolment code and email address
```
benjefferies, line 34: I wonder if we could remove the need for this documentation if we refactored out the specific logic into those bit sized chunks of logic. Somethign like 
```
def post_responsdent(party, tran, session):
  _validate(...)
  _register_responsdent(..)
  find_case(...)
.....
```

andrewmil (approved):
Changes look good, the structure looks better now.
I started making comments about more fine tuned refactoring but that's probably for more changes down the line

ras_party/controllers/account_controller.py
```diff
@@ -0,0 +1,461 @@
...
+from ras_party.support.verification import decode_email_token
+
+log = get_logger()
+
+NO_RESPONDENT_FOR_PARTY_ID = 'There is no respondent with that party ID '
+EMAIL_VERIFICATION_SENT = 'A new verification email has been sent'
+
+
+@translate_exceptions
+@transactional
```
benjefferies: I wonder if it would be simplified if each flask request started a transaction, committed or rollbacked based on success or not?

ras_party/controllers/account_controller.py
andrewmil: This references the other comment I mentioned in controllers/queries.py

ras_party/controllers/queries.py
benjefferies: I agree, I was going to mention this in another PR but held myself back. I think anywhere we expect to get only one item returned get is more descriptive

ras_party/views/account_view.py
```diff
@@ -0,0 +1,82 @@
...
+def get_pw(username):
+    config_username = current_app.config['SECURITY_USER_NAME']
+    config_password = current_app.config['SECURITY_USER_PASSWORD']
+    if username == config_username:
+        return config_password
+
+
+@account_view.route('/respondents/change_email', methods=['PUT'])
+@account_view.route('/respondents/email', methods=['PUT'])
+@auth.login_required
```
benjefferies: If we expect to `@auth.login_required` for every api in this blueprint we might be able to use http://flask.pocoo.org/docs/0.12/api/#flask.Flask.before_request. Same with `@log_route`

benjefferies (changes requested):
Some awesome improvements here!

ras_party/controllers/queries.py
tej99: 'get_business_by_ref' and 'get_business_by_id' already exists as business controllers

ras_party/controllers/account_controller.py
tej99: Not sure it's worth splitting out as of now, nothing else currently uses this functionality.

ras_party/views/account_view.py
```diff
@@ -13,6 +13,13 @@
 auth = HTTPBasicAuth()
 
 
+@account_view.before_request
```
andrewmil: Is there a reason to do this for each blueprint rather than just a general `@app.before_request` for all incoming requests?

benjefferies (approved):

ras_party/views/account_view.py
tej99: Not all blueprints use @auth.login_required before request

PR #68, opened 2017-10-27
WIP [TBL111] Use pipenv for deterministic builds

benjefferies:
Pipenv enables us to lock down and guarantee what dependencies we get
when we install dependencies. This works by using the pip hashes when
running pipenv install --dev. The hashes are stored in a generated
file call Pipfile.lock.

We no longer need requirement files instead pipenv uses a file called
Pipfile. The documentation can be found
https://github.com/kennethreitz/pipenv.


benjefferies:
Currently investigating why pipenv installs dev packages https://github.com/cloudfoundry/python-buildpack/issues/88

PR #69, opened 2017-11-01
[tbl111] Use pipenv for deterministic builds

benjefferies:
Pipenv enables us to lock down and guarantee what dependencies we get
when we install dependencies. This works by using the pip hashes when
running pipenv install --dev. The hashes are stored in a generated
file call Pipfile.lock.

We no longer need requirement files instead pipenv uses a file called
Pipfile. The documentation can be found
https://github.com/kennethreitz/pipenv.


jwgwalton (approved):
Clear and concise README.

jcox-dev (approved):

benjefferies:
Deployed to Int and tested

PR #71, opened 2017-11-13
Add instructions to run app locally to README

gemmai95:
Update README to include instructions for running service locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,10 +22,18 @@ pipenv install
 ```
 
 ## Running
+
 [Install Docker](https://docs.docker.com/engine/installation/)
 ```bash
 docker-compose up
 ```
+
+or run locally (this requires postgres running locally on port 5432):
+```bash
+pipenv run python3 run.py
+```
+
+
 
 To test the service is up:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -22,11 +22,19 @@ pipenv install
 ```
 
 ## Running
+
 [Install Docker](https://docs.docker.com/engine/installation/)
 ```bash
 docker-compose up
 ```
 
+or run locally (this requires postgres running locally on port 5432):
```
andrewmil, line 31: From some testing locally you also need to change the `ras-party-db.uri` to be `"postgres://postgres:postgres@localhost:5432/postgres"`

andrewmil: Or that should just be the new default maybe?

gemmai95: I changed it in the config file to localhost locally. I discussed this with @davidmort and he wanted to leave it with him for now.

andrewmil (approved):
Looks good if we know what we're doing about the database uri

tej99 (approved):

PR #73, opened 2017-11-27
Retry db connection on failure

benjefferies:
Retry database connection on DatabaseError every 2s for up to 30s. This
allows the app to be more robust when starting up. For instance when
running up apps via docker-compose the app should be able to recover if
the database is not up first.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,8 +4,11 @@ verify_ssl = true
 name = "pypi"
 url = "https://pypi.python.org/simple"
 
+
 [requires]
+
 python_version = "3.6"
+
 
 [packages]
 
@@ -24,13 +27,14 @@ requests = "==2.17.3"
 structlog = "==17.2.0"
 flask-httpauth = "==3.2.3"
 ras-common-utils = {git = "https://github.com/ONSdigital/ras-common-utils.git", ref = "0.0.7"}
+retrying = "*"
 
 
 [dev-packages]
 
 tox = "*"
 coverage = "*"
-flake8 = "*"
+"flake8" = "*"
 flask-testing = "*"
 pytest-cov = "*"
 setuptools = "==32.1.0"
--- a/run.py
+++ b/run.py
@@ -1,12 +1,15 @@
 from json import loads
 
+import structlog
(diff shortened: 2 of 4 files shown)


tej99 (approved):
LGTM. Out of curiosity, what's the advantage of using the retrying package instead of a for loop with time delay?

benjefferies:
@tej99 Less code, the code has already been tested

Pipfile
```diff
@@ -24,13 +27,14 @@ requests = "==2.17.3"
...
 ras-common-utils = {git = "https://github.com/ONSdigital/ras-common-utils.git", ref = "0.0.7"}
+retrying = "*"
 
 
 [dev-packages]
 
 tox = "*"
 coverage = "*"
-flake8 = "*"
+"flake8" = "*"
```
JamesGardiner, line 37: why the quotes?

run.py
```diff
@@ -30,6 +32,11 @@ def create_app(config):
     return app
 
 
+def retry_if_database_error(exception):
+    return isinstance(exception, DatabaseError)
+
+
+@retry(retry_on_exception=retry_if_database_error, wait_fixed=2000, stop_max_delay=30000)
```
JamesGardiner: Do you want to gracefully handle the `RetryError` that is raised?

JamesGardiner: https://github.com/rholder/retrying/blob/1d5699348d707e377aad7488da6a8a1b48a65933/retrying.py#L286

Pipfile
benjefferies: It was automatically added when pipenv install. I can remove

run.py
benjefferies: I guess the only way we can handle it is by logging the error as at that point it seems irrecoverable?

Pipfile
jcox-dev: Leave it in. It's Pipenv normalising the package names. Those with numbers or special characters are intentionally quoted probably from checking against PyPI real names.

jcox-dev (approved):

JamesGardiner (approved):

PR #79, opened 2017-12-08
party versioning

tej99:
-data model changes for party versioning 
-add alembic to manage party versioning
-returning sampleSummaryId in json response for business/party details


alembic/versions/92464d9a418f_set_up_database.py
```diff
@@ -0,0 +1,39 @@
...
+
+    sql_query = "INSERT INTO partysvc.business_attributes (business_id, attributes) " \
+                "SELECT party_uuid, attributes FROM partysvc.business"
+    conn.execute(sql_query)
+
+    op.drop_column('business', 'attributes', schema='partysvc')
+
+
+def downgrade():
+    pass
```
rjweeks: Should we include the downgrade?

rjweeks (approved):

Pipfile
```diff
@@ -28,7 +28,7 @@ structlog = "==17.2.0"
 flask-httpauth = "==3.2.3"
 ras-common-utils = {git = "https://github.com/ONSdigital/ras-common-utils.git", ref = "0.0.7"}
 retrying = "*"
-
+alembic = "*"
```
emilioward: Does this need to be a more specific version of alembic?

emilioward (approved):
Looks good tests pass

JamesGardiner (approved):

jcox-dev (changes requested):
Do we need a few simple steps added to the README mentioning that the repo uses alembic?

migrations/README
```diff
@@ -1 +1,14 @@
...
+Upgrade to the most recent version
+```
+alembic upgrade head
+```
+
+
+## Downgrade
+Downgrade to a particular version
+```
+alembic upgrade {version}
```
jcox-dev: "Or to undo all migrations: `alembic downgrade base`" might be worth adding in.

jcox-dev (changes requested):

jcox-dev (approved):

migrations/env.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+# this is the Alembic Config object, which provides
+# access to the values within the .ini file in use.
+config = context.config
+
+# override sqlalchemy.url
+config_path = 'config/config.yaml'
+with open(config_path, 'r') as f:
+    config_data = yaml.load(f)
```
andrewmil, line 21: this should be `yaml.safe_load(f)`

Using `yaml.load` potentially allows arbirtary code execution 
https://security.openstack.org/guidelines/dg_avoid-dangerous-input-parsing-libraries.html

andrewmil (changes requested):
We shouldn't use `yaml.load()`

migrations/env.py
tej99: We won't be using yaml. Once @jcox-dev's config refactor goes in, so i'll probably just leave this in for now.

andrewmil (approved):

PR #80, opened 2017-12-12
Add business sample link endpoint

tej99:
Add endpoint to update link between business attributes versions and a collection exercise required for party versioning.

### DIFF
--- a/ras_party/controllers/business_controller.py
+++ b/ras_party/controllers/business_controller.py
@@ -86,20 +86,20 @@ def businesses_post(business_data, session):
 
 
 @with_db_session
-def businesses_sample_ce_link(sample, link_data, session):
+def businesses_sample_ce_link(sample, ce_data, session):
     """
     Update business versions to have the correct collection exercise and sample link.
 
     :param sample: the sample summary id to update.
-    :param link_data: dictionary containing the collectionExerciseId to link with sample.
+    :param ce_data: dictionary containing the collectionExerciseId to link with sample.
     :param session: database session.
     """
 
     v = Validator(Exists('collectionExerciseId'))
-    if not v.validate(link_data):
+    if not v.validate(ce_data):
         raise RasError(v.errors, 400)
 
-    collection_exercise_id = link_data['collectionExerciseId']
+    collection_exercise_id = ce_data['collectionExerciseId']
 
     session.query(BusinessAttributes).filter(BusinessAttributes.sample_summary_id == sample)\
         .update({'collection_exercise': collection_exercise_id})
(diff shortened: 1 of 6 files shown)


ras_party/controllers/business_controller.py
```diff
@@ -83,3 +83,23 @@ def businesses_post(business_data, session):
...
+def businesses_sample_ce_link(sample, link_data, session):
+    """
+    Update business versions to have the correct collection exercise and sample link.
+
+    :param sample: the sample summary id to update.
+    :param link_data: dictionary containing the collectionExerciseId to link with sample.
+    :param session: database session.
+    """
+
+    v = Validator(Exists('collectionExerciseId'))
```
andrewmil: Should we do this validation in the controller? I feel like this is something we should do before passing it from the view. This would remove the need to have the default value for payload in the view

ras_party/views/business_view.py
```diff
@@ -51,3 +51,12 @@ def get_business_by_ref(ref):
...
     response = ras_party.controllers.business_controller.get_business_by_ref(ref, verbose=verbose)
     return make_response(jsonify(response), 200)
+
+
+@business_view.route('/businesses/sample/link/<sample>', methods=['PUT'])
+def put_business_attributes_ce(sample):
+    payload = request.get_json() or {}
+    ras_party.controllers.business_controller.businesses_sample_ce_link(sample, payload)
+    response = payload
+    response['sampleSummaryId'] = sample
```
andrewmil: We can just write `response = {**payload, "sampleSummaryId": sample}` here

ras_party/controllers/business_controller.py
```diff
@@ -83,3 +83,23 @@ def businesses_post(business_data, session):
     b = Business.from_party_dict(party_data)
     session.merge(b)
     return b.to_business_dict()
+
+
+@with_db_session
+def businesses_sample_ce_link(sample, link_data, session):
```
andrewmil: Is there a better name than link_data than we could use here?

ras_party/controllers/business_controller.py
andrewmil: If link data only has collectionExerciseId in it then we should probably validate/parse this in the view then just pass the collection exercise to this function directly. This would make it clearer what this function is doing

andrewmil (changes requested):
A few small comments, functionally everything looks good.

I think that parsing/validating the incoming data should be separate from the actions we're trying to perform

ras_party/controllers/business_controller.py
tej99: I agree but I was just following the same pattern all the other endpoints have. Maybe have a refactor of all the controllers that implement the validation in the controller as part of another ticket if it is a big issue.

ras_party/views/business_view.py
tej99: Could do.

andrewmil (approved):

rjweeks (approved):

PR #81, opened 2017-12-13
Remove use of ras-common-utils and improve use of config

jcox-dev:
Removes dependency on ras-common-utils and use of yaml config (12FA).


jcox-dev:
[Remove use of common-utils](https://trello.com/c/10pkGUX9/3-remove-use-of-common-utils)

config.py
```diff
@@ -0,0 +1,146 @@
...
+
+    # dependencies
+
+    RAS_PARTY_DATABASE_SCHEMA = os.getenv('RAS_PARTY_DATABASE_SCHEMA', 'partysvc')
+    RAS_PARTY_DATABASE_URI = os.getenv('RAS_PARTY_DATABASE_URI', 'postgres://postgres:postgres@localhost:5432/postgres')
+
+    RAS_PUBLIC_WEBSITE_PROTOCOL = os.getenv('RAS_OAUTH_SERVICE_PROTOCOL', 'http')
+    RAS_PUBLIC_WEBSITE_HOST = os.getenv('RAS_OAUTH_SERVICE_HOST', 'dummy.ons.gov.uk')
+    RAS_PUBLIC_WEBSITE_PORT = os.getenv('RAS_OAUTH_SERVICE_PORT', 80)
+    RAS_PUBLIC_WEBSITE_URL = f'{RAS_PUBLIC_WEBSITE_PROTOCOL}://{RAS_PUBLIC_WEBSITE_HOST}:{RAS_PUBLIC_WEBSITE_PORT}'
```
benjefferies: I think we should look at collapsing these down going forward into just the URL env vars

jcox-dev: I agree. Would make more sense to parse a single URL if the component parts are required.

config.py
```diff
@@ -0,0 +1,122 @@
...
+    SECURITY_USER_NAME = os.getenv('SECURITY_USER_NAME', 'dummy_user')
+    SECURITY_USER_PASSWORD = os.getenv('SECURITY_USER_PASSWORD', 'dummy_password')
+    SECURITY_REALM = os.getenv('SECURITY_REALM', 'sdc')
+
+    # dependencies
+
+    RAS_PARTY_DATABASE_SCHEMA = os.getenv('RAS_PARTY_DATABASE_SCHEMA', 'partysvc')
+    RAS_PARTY_DATABASE_URI = os.getenv('RAS_PARTY_DATABASE_URI', 'postgres://postgres:postgres@localhost:5432/postgres')
+
+    RAS_PUBLIC_WEBSITE_PROTOCOL = os.getenv('RAS_OAUTH_SERVICE_PROTOCOL', 'http')
```
andrewmil: `RAS_PUBLIC_WEBSITE_PROTOCOL` should be set to it's equivalent environment variable, Not `RAS_OAUTH_SERVICE`.
Same with the two lines below

ras_party/clients/oauth_client.py
```diff
@@ -3,14 +3,13 @@
 
 class OauthClient:
     def __init__(self, config):
-        self.service_config = config.dependency['oauth2-service']
-        self.client_id = self.service_config['client_id']
-        self.client_secret = self.service_config['client_secret']
+        self.service = config['RAS_OAUTH_SERVICE']
```
andrewmil: This is leftover from what we've previously been doing but I don't think it's good to be passing the config object around. Config should be set from`(current_)app.config[...]`

ras_party/support/session_decorator.py
```diff
@@ -1,10 +1,13 @@
...
 from flask import current_app
-from ras_common_utils.ras_error.ras_error import RasError, RasDatabaseError
-from structlog import get_logger
+import structlog
 
-log = get_logger()
+from ras_party.exceptions import RasError, RasDatabaseError
+
+
+logger = structlog.wrap_logger(logging.getLogger(__name__))
```
andrewmil, line 10: In some files we're calling this logger and in others we're calling it log, should we be more consistent?

jcox-dev:
[Make party service 12FA compliant](https://trello.com/c/Wd0IK72m/18-make-party-service-12fa-compliant)

Pipfile
```diff
@@ -22,11 +22,9 @@ sqlalchemy = "==1.1.10"
 "psycopg2" = "==2.7.1"
 pycrypto = "==2.6.1"
 python-dateutil = "==2.6.0"
-pyyaml = "==3.12"
 requests = "==2.17.3"
 structlog = "==17.2.0"
 flask-httpauth = "==3.2.3"
-ras-common-utils = {git = "https://github.com/ONSdigital/ras-common-utils.git", ref = "0.0.7"}
 retrying = "*"
```
andrewmil: Should we specify the version of this as well?

andrewmil (changes requested):
Looks good, removing common-utils makes the way we initialise the app a lot clearer.

This change would allow us to change the structure of the app to something more standard for flask. e.g. the app object should be created inside the ras_party folder rather than at the top level

config.py
jcox-dev: Knew that copy-pasta would bite me eventually...

ras_party/support/session_decorator.py
jcox-dev: Always. I thought I'd caught all the different ones.

andrewmil (approved):

run.py
```diff
@@ -1,21 +1,26 @@
...
 
-def create_app(config):
+logger = structlog.wrap_logger(logging.getLogger(__name__))
+
+
+def create_app(config=None):
     # create and configure the Flask app
     app = Flask(__name__)
-    app.config.from_ras_config(config)
+    app_config = f"config.{config or os.environ.get('APP_SETTINGS', 'Config')}"
```
emilioward, line 22: is that "f" supposed to be there?

jcox-dev: Yep. It's an [f-string](https://www.python.org/dev/peps/pep-0498/) for formatting the config string in place.

app.py
```diff
@@ -1,26 +1,24 @@
...
 """
 This is a duplicate of run.py, with minor modifications to support gunicorn execution.
 """
 
-logger = structlog.get_logger()
+logger = structlog.wrap_logger(logging.getLogger(__name__))
 
-
-config_path = 'config/config.yaml'
-config = ras_config.from_yaml_file(config_path)
```
tej99, line 18: The 'from_yaml_file' also has logic which checks if there is a VCAP Env var which overrides the config file. this would break in the cloudfoundry environment because the database will not be configured properly.

This is the code from [ras_common_utils](https://github.com/ONSdigital/ras-common-utils/blob/master/ras_common_utils/ras_config/ras_config.py):
 ```python
def from_yaml_file(path):
    with open(path) as f:
        data = yaml.load(f.read())
    return make(data)

def make(config_data):
    vcap_application = getenv('VCAP_APPLICATION')
    if vcap_application:
        log.info("CloudFoundry detected. Creating CloudFoundry configuration object.")
        return RasCloudFoundryConfig(config_data)
    else:
        log.info("CloudFoundry not detected. Creating standard configuration object.")
        return RasConfig(config_data)
```

File: migrations/env.py [hunk 8]
```diff
@@ -1,25 +1,22 @@
...
 
 # override sqlalchemy.url
-config_path = 'config/config.yaml'
-with open(config_path, 'r') as f:
-    config_data = yaml.load(f)
-config.set_main_option("sqlalchemy.url", config_data['dependencies']['ras-party-db']['uri'])
+from config import Config # NOQA
+db_uri = Config.RAS_PARTY_DATABASE_URI
+if db_uri:
+    config.set_main_option("sqlalchemy.url", db_uri)
```
tej99: This was a bug introduced by me. Again the same situation as before, the database uri is taken from the config file (which is correct for dev environments) but it needs to get overridden by the VCAP env if it's there.

tej99 (changes requested):
Looks gd, but looks like it's missing the logic to configure if it's in the cloudfoundry env.

app.py
andrewmil: Looks like there's a lot of overcomplicated stuff going on to load from VCAP in common_utils.

This should be simple with the cfenv package. We're using it to [get redis config in frontstage](https://github.com/ONSdigital/ras-frontstage/blob/master/frontstage/cloud/cloudfoundry.py)

config.py
```diff
@@ -13,8 +18,15 @@ class Config(object):
     LOGGING_LEVEL = os.getenv('LOGGING_LEVEL', 'INFO')
     SECRET_KEY = os.getenv('SECRET_KEY', 'aardvark')
     EMAIL_TOKEN_SALT = os.getenv('EMAIL_TOKEN_SALT', 'aardvark')
-    DATABASE_SCHEMA = os.getenv('DATABASE_SCHEMA', 'ras-party')
-    DATABASE_URI = os.getenv('DATABASE_URI', "sqlite:///:memory:")
+    PARTY_SCHEMA = os.getenv('PARTY_SCHEMA', 'ras_party/schemas/party_schema.json')
+
+    if cf.detected:
```
andrewmil: 👍  Doing this inside the config object makes a lot of sense

tej99 (approved):
Awesome 👍

config.py
```diff
@@ -0,0 +1,134 @@
...
+    LOGGING_LEVEL = os.getenv('LOGGING_LEVEL', 'INFO')
+    SECRET_KEY = os.getenv('SECRET_KEY', 'aardvark')
+    EMAIL_TOKEN_SALT = os.getenv('EMAIL_TOKEN_SALT', 'aardvark')
+    PARTY_SCHEMA = os.getenv('PARTY_SCHEMA', 'ras_party/schemas/party_schema.json')
+
+    if cf.detected:
+        DATABASE_SCHEMA = 'partysvc'
+        DATABASE_URI = cf.db.credentials['uri']
+    else:
+        DATABASE_SCHEMA = os.getenv('DATABASE_SCHEMA', 'partysvc')
```
benjefferies, line 27: Does the schema need to be configurable?

config.py
```diff
@@ -0,0 +1,134 @@
...
+
+    REQUESTS_GET_TIMEOUT = os.getenv('REQUESTS_GET_TIMEOUT', 20)
+    REQUESTS_POST_TIMEOUT = os.getenv('REQUESTS_POST_TIMEOUT', 20)
+    SECURITY_USER_NAME = os.getenv('SECURITY_USER_NAME', 'dummy_user')
+    SECURITY_USER_PASSWORD = os.getenv('SECURITY_USER_PASSWORD', 'dummy_password')
+    SECURITY_REALM = os.getenv('SECURITY_REALM', 'sdc')
+
+    # dependencies
+
+    RAS_PARTY_DATABASE_SCHEMA = os.getenv('RAS_PARTY_DATABASE_SCHEMA', 'partysvc')
```
benjefferies: Duplicate env with a different name - https://github.com/ONSdigital/ras-party/pull/81/files#diff-fdf313ed2303059c44a9ebfffbef8eebR27?

config.py
```diff
@@ -0,0 +1,134 @@
...
+    RAS_PARTY_DATABASE_URI = os.getenv('RAS_PARTY_DATABASE_URI', 'postgres://postgres:postgres@localhost:5432/postgres')
```
benjefferies: Duplicate env with a different name - https://github.com/ONSdigital/ras-party/pull/81/files#diff-fdf313ed2303059c44a9ebfffbef8eebR28?

run.py
```diff
@@ -35,29 +40,51 @@ def create_app(config):
...
 
 
+def create_database(db_connection, db_schema):
+    from ras_party.models import models
+
+    def current_request():
+        return _app_ctx_stack.__ident_func__()
+
+    engine = create_engine(db_connection, convert_unicode=True)
+    session = scoped_session(sessionmaker(), scopefunc=current_request)
```
benjefferies, line 50: If we're using scoped_session do we need the `with_db_session` stuff?

tox.ini
```diff
@@ -3,11 +3,9 @@ envlist = py36
 
 [testenv]
 passenv=HOME
-deps = pipenv
 commands=
-    pipenv install --dev
-    pipenv run flake8 .
-    pipenv run py.test --cov=ras_party --cov-report html --cov-report term-missing []
+    flake8 .
```
benjefferies, line 7: Good shout to get rid of the `pipenv install` stuff from tox

benjefferies (changes requested):
Looks

config.py
jcox-dev: I'm not sure myself - probably not with our current deployment.

run.py
jcox-dev: Can. Of. Worms. But yeah, probably beyond this PR though.

benjefferies (approved):

PR #82, opened 2017-12-15
use ras common utils from pipy

tej99:
update pipfile to use ras-common-utils from pypi


rhiGriff (approved):

jcox-dev (changes requested):
Will this be necessary once #81 is merged?

tej99:
@jcox-dev No but we need this change in right now

jcox-dev (approved):

PR #85, opened 2017-12-20
Automate alembic commands

tej99:
This gets rid of the need to run alembic commands through all the environments.


gemmai95 (approved):

jcox-dev (changes requested):
Testing against db created from master:

```
$ cat .env
APP_SETTINGS=DevelopmentConfig
DATABASE_URI="postgres://postgres:postgres@localhost:5432/postgres"
DATABASE_SCHEMA="partysvc"
$ pipenv shell
$ git checkout master
$ docker-compose up -d db
$ python run.py (ctrl+c)
$ git checkout automate-alembic-commands
$ python run.py
...
Running upgrade  -> 92464d9a418f
{"event": "(psycopg2.ProgrammingError) relation \"business_attributes\" already exists\n [SQL: '\\nCREATE TABLE partysvc.business_attributes (\\n\\tid SERIAL NOT NULL, \\n\\tbusiness_id UUID, \\n\\tsample_summary_id TEXT, \\n\\tcollection_exercise TEXT, \\n\\tattributes JSONB, \\n\\tcreated_on TIMESTAMP WITHOUT TIME ZONE, \\n\\tPRIMARY KEY (id), \\n\\tFOREIGN KEY(business_id) REFERENCES partysvc.business (party_uuid)\\n)\\n\\n']", "level": "error", "service": "ras-party", "created_at": "2017-12-20T15:071513782470"}
```

run.py
```diff
@@ -52,17 +56,37 @@ def current_request():
     # TODO: change this
     engine.session = session
 
-    # fix-up the postgres schema:
+    alembic_cfg = Config("alembic.ini")
+    alembic_cfg.attributes['configure_logger'] = False
+
+    logger.info(f"Creating database with uri '{db_connection}'")
```
rhiGriff: Is there a reason we've moved away from the SDC logging standards?

tej99:
@jcox-dev  The reason behind that is because the master branch has the updated models so it creates the 'business_attributes' table when bringing up the application. If you manually delete the 'business_attributes' table and bring up the application using this branch it will do the data migration correctly.

run.py
tej99: Not sure that log line was already there. Is this against sdc logging standards?

rhiGriff: https://collaborate2.ons.gov.uk/confluence/display/SDC/Logging

emilioward: That was something @jcox-dev added in the party service refactor

benjefferies (changes requested):
Looks good but we should probably have some unit tests to check the logic

jcox-dev (approved):

benjefferies (approved):

PR #88, opened 2017-12-22
Split logic for loading new/existing businesses

tej99:
Fixing a bug where uploading an existing business removed all previous business_attributes relations.

### DIFF
--- a/ras_party/controllers/business_controller.py
+++ b/ras_party/controllers/business_controller.py
@@ -71,13 +71,15 @@ def businesses_post(business_data, session):
     if errors:
         raise RasError([e.split('\n')[0] for e in errors], status_code=400)
 
-    existing_business = query_business_by_ref(party_data['sampleUnitRef'], session)
-    if existing_business:
-        party_data['id'] = str(existing_business.party_uuid)
-
-    b = Business.from_party_dict(party_data)
-    session.merge(b)
-    return b.to_business_dict()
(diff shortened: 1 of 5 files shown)


ras_party/models/models.py
```diff
@@ -90,18 +89,25 @@ def _get_respondents_associations(respondents):
             associations.append(respondent_dict)
         return associations
 
+    def add_versioned_attributes(self, party):
```
jcox-dev, line 92: We should have a test for this new method.

ras_party/controllers/business_controller.py
```diff
@@ -71,13 +71,15 @@ def businesses_post(business_data, session):
...
 
-    existing_business = query_business_by_ref(party_data['sampleUnitRef'], session)
-    if existing_business:
-        party_data['id'] = str(existing_business.party_uuid)
-
-    b = Business.from_party_dict(party_data)
-    session.merge(b)
-    return b.to_business_dict()
+    business = query_business_by_ref(party_data['sampleUnitRef'], session)
+    if business:
```
jcox-dev, line 75: This is a lot clearer 👍

jcox-dev (changes requested):
Looks good but probably need to add tests for the new method.

jcox-dev (approved):

andrewmil (approved):
Tested by hitting endpoints locally, everything looks good :+1:

PR #96, opened 2018-02-01
Update get attributes function

andrewmil:
Update the method party uses to find business attributes by collection exercise so that when no matching collection exercise is found, the most recent attributes are returned.

This doesn't fix any broken changes, just makes party service more resilient if it has bad data


emilioward (approved):
👍

tej99 (approved):

PR #101, opened 2018-02-06
Disable IAC code when survey added

LJBabbage:
The IAC should have its state change to active false when a new survey is added. This is not the long term intended solution and should be done in case. This is just for release 15 and should be reviewed with that in mind

### DIFF
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -389,6 +389,12 @@ def add_new_survey_for_respondent(payload, tran, session):
 
     post_case_event(str(case_id), str(respondent_party_id), "RESPONDENT_ENROLED", "Respondent enroled")
 
+    # TODO This is not the correct way to do this and should be done in the case service.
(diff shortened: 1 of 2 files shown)


gemmai95 (approved):

ras_party/controllers/account_controller.py
```diff
@@ -389,6 +389,12 @@ def add_new_survey_for_respondent(payload, tran, session):
 
     post_case_event(str(case_id), str(respondent_party_id), "RESPONDENT_ENROLED", "Respondent enroled")
 
+    # TODO This is not the correct way to do this and should be done in the case service.
+    # TODO This is very much temporary and should only be for release 15
+    iac_code = disable_iac_code(enrolment_code)
+    if iac_code.get('active', True):
+        raise RasError("Enrolment code has not been disabled", status=400)
```
benjefferies: We might want to propagate the the response code returned from the iac service here

benjefferies (approved):

PR #103, opened 2018-02-13
allow key words in diffrent order

tej99:
Searching 'castle ltd' should return 'castle heights ltd'


gemmai95 (approved):

tej99:
[US043: INT Find an RU [Large]](https://trello.com/c/WyRWFKXq/38-us043-int-find-an-ru-large)

emilioward (approved):

PR #105, opened 2018-03-07
Us057 change respondent name and number

emilioward:
This PR is linked to the other Us057 requests in backstage (https://github.com/ONSdigital/ras-backstage/pull/73) and response operations (awaiting PR for r-ops).

This adds in the functionality to update a respondents `first name`, `last_name`, `telephone`

### DIFF
--- a/ras_party/controllers/queries.py
+++ b/ras_party/controllers/queries.py
@@ -81,10 +81,10 @@ def update_respondent_details(respondent_data, session):
 
     logger.debug('Updating respondent details', respondent_id=respondent_data['respondent_id'])
 
-    return session.query(Respondent).filter(Respondent.party_uuid == respondent_data['respondent_id']).update({
-                                            Respondent.first_name: respondent_data['firstName'],
-                                            Respondent.last_name: respondent_data['lastName'],
-                                            Respondent.telephone: respondent_data['telephone']})
+    session.query(Respondent).filter(Respondent.party_uuid == respondent_data['respondent_id']).update({
+                                     Respondent.first_name: respondent_data['firstName'],
+                                     Respondent.last_name: respondent_data['lastName'],
+                                     Respondent.telephone: respondent_data['telephone']})
 
 
 def search_businesses(search_query, session):
--- a/ras_party/controllers/queries.py
+++ b/ras_party/controllers/queries.py
@@ -79,7 +79,7 @@ def query_change_respondent_details(respondent_data, session):
     :return: updated respondent details
     """
 
-    logger.debug('Changing respondent details', respondent_data=respondent_data)
+    logger.debug('Changing respondent details', respondent_id=respondent_data['respondent_id'])
 
     return session.query(Respondent).filter(Respondent.party_uuid == respondent_data['respondent_id']).update({
                                             Respondent.first_name: respondent_data['firstName'],
--- a/ras_party/controllers/queries.py
+++ b/ras_party/controllers/queries.py
@@ -67,7 +67,7 @@ def query_business_respondent_by_respondent_id_and_business_id(business_id, resp
     return response
 
 
-def update_respondent_details(respondent_data, session):
+def update_respondent_details(respondent_data, respondent_id, session):
     """
     Query to return respondent, respondent_data consists of the following parameters
     :param respondent_data:
@@ -79,9 +79,9 @@ def update_respondent_details(respondent_data, session):
     :return: updated respondent details
     """
 
-    logger.debug('Updating respondent details', respondent_id=respondent_data['respondent_id'])
+    logger.debug('Updating respondent details', respondent_id=respondent_id)
 
-    session.query(Respondent).filter(Respondent.party_uuid == respondent_data['respondent_id']).update({
+    session.query(Respondent).filter(Respondent.party_uuid == respondent_id).update({
                                      Respondent.first_name: respondent_data['firstName'],
                                      Respondent.last_name: respondent_data['lastName'],
                                      Respondent.telephone: respondent_data['telephone']})
--- a/ras_party/controllers/queries.py
+++ b/ras_party/controllers/queries.py
@@ -67,9 +67,10 @@ def query_business_respondent_by_respondent_id_and_business_id(business_id, resp
     return response
 
 
-def query_change_respondent_details(respondent_first_name, respondent_last_name, respondent_tel_number, session):
+def query_change_respondent_details(id, respondent_first_name, respondent_last_name, respondent_tel_number, session):
     """
     Query to return respondent
+    :param id:
     :param respondent_first_name:
     :param respondent_last_name:
     :param respondent_tel_number:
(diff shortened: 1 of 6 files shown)


ras_party/controllers/queries.py
```diff
@@ -67,6 +67,26 @@ def query_business_respondent_by_respondent_id_and_business_id(business_id, resp
...
+    :param respondent_data:
+        respondent_id: id of the respondent
+        first_name:
+        last_name:
+        telephone:
+    :param session
+    :return: updated respondent details
+    """
+
+    logger.debug('Changing respondent details', respondent_data=respondent_data)
```
gemmai95: The respondent data should not be being logged out. Only their id.

gemmai95 (changes requested):

emilioward:
[US057: INT Change Respondent Name & Number [M]](https://trello.com/c/1ZGcBo4W/45-us057-int-change-respondent-name-number-m)

ras_party/controllers/queries.py
```diff
@@ -67,6 +67,26 @@ def query_business_respondent_by_respondent_id_and_business_id(business_id, resp
     return response
 
 
+def query_change_respondent_details(respondent_data, session):
```
benjefferies: I think the method name update_respondent_details is good enough

ras_party/controllers/respondent_controller.py
```diff
@@ -40,3 +41,17 @@ def get_respondent_by_email(email, session):
...
+@with_db_session
+def change_respondent_details(respondent_data, session):
+    """
+    :param respondent_data:
+    :param session:
+    :return:
+    """
+    v = Validator(IsUuid('respondent_id'))
+    if not v.validate({'respondent_id': respondent_data['respondent_id']}):
+        raise RasError(v.errors, status=404)
```
benjefferies: I think this should be a 400 if the id is invalid

test/test_respondent_controller.py
```diff
@@ -101,6 +101,26 @@ def test_get_respondent_by_email_returns_correct_representation(self):
...
 
+    def test_change_respondent_details_success(self):
+        self.populate_with_respondent()
+        payload = {
+            "respondent_id": "438df969-7c9c-4cd4-a89b-ac88cf0bfdf3",
+            "firstName": "John",
+            "lastName": "Snow",
+            "telephone": "07837230942"
+        }
+        self.change_respondent_details(payload, 200)
```
benjefferies: Not keen on this approach of asserting. It's not very clear that change_respondent_details is doing the assertion by the name. I also think it's more readable if the assertion is done in the test method

ras_party/controllers/queries.py
benjefferies: I think the method name update_respondent_details is good enough?

benjefferies (changes requested):
Looks good, few minor points

ras_party/views/respondent_view.py
```diff
@@ -38,3 +38,10 @@ def get_respondent_by_id(id):
 def get_respondent_by_email(email):
     response = ras_party.controllers.respondent_controller.get_respondent_by_email(email)
     return make_response(jsonify(response), 200)
+
+
+@respondent_view.route('/respondents/change_respondent_details', methods=['PUT'])
```
andrewmil: I feel like the endpoint should be `/respondents/change_respondent_details/<respondent_id>` with `respondent_id` in the url rather than the payload.

ras_party/controllers/respondent_controller.py
andrewmil: Yeah agreed. This is a check to see if the format of the `respondent_id` is correct, not to check if it exists

ras_party/controllers/queries.py
andrewmil: I prefer this to having the work query in the name as well, the word query is in the name of several other functions in this file however

andrewmil (changes requested):
Looks good, just a few small comments

ras_party/controllers/queries.py
```diff
@@ -67,6 +67,26 @@ def query_business_respondent_by_respondent_id_and_business_id(business_id, resp
...
+        first_name:
+        last_name:
+        telephone:
+    :param session
+    :return: updated respondent details
+    """
+
+    logger.debug('Updating respondent details', respondent_id=respondent_data['respondent_id'])
+
+    return session.query(Respondent).filter(Respondent.party_uuid == respondent_data['respondent_id']).update({
```
benjefferies: Don't think we even need to return this to be honest?

andrewmil (approved):

ras_party/controllers/queries.py
andrewmil: If the query fails this will throw an exception right? If that's the case then we shouldn't have to return anything

gemmai95 (approved):

ras_party/views/respondent_view.py
davidmort: yes def agree with this

test/test_respondent_controller.py
```diff
@@ -101,6 +101,26 @@ def test_get_respondent_by_email_returns_correct_representation(self):
...
 
+    def test_update_respondent_details_success(self):
+        self.populate_with_respondent(respondent=self.mock_respondent_with_id)
+        respondent_id = self.mock_respondent_with_id['id']
+        payload = {
+            "firstName": "John",
+            "lastName": "Snow",
+            "telephone": "07837230942"
+            }
+        self.change_respondent_details(respondent_id, payload, 200)
```
benjefferies, line 112: Can we do the assertion here?

Something like 

```
resp = self.change_respondent_details(respondent_id, payload)
assert resp.status_code == 200
```

benjefferies (approved):

gemmai95 (changes requested):
The API.md needs to be updated with the new endpoint.

gemmai95 (approved):

ras_party/controllers/queries.py
```diff
@@ -67,6 +67,26 @@ def query_business_respondent_by_respondent_id_and_business_id(business_id, resp
...
+def update_respondent_details(respondent_data, respondent_id, session):
+    """
+    Query to return respondent, respondent_data consists of the following parameters
+    :param respondent_data:
+        respondent_id: id of the respondent
+        first_name:
+        last_name:
+        telephone:
+    :param session
+    :return: updated respondent details
```
jcox-dev: Docstring should be updated to reflect that this doesn't return anything.

ras_party/views/respondent_view.py
```diff
@@ -38,3 +38,10 @@ def get_respondent_by_id(id):
 def get_respondent_by_email(email):
     response = ras_party.controllers.respondent_controller.get_respondent_by_email(email)
     return make_response(jsonify(response), 200)
+
+
+@respondent_view.route('/respondents/change_respondent_details/<respondent_id>', methods=['PUT'])
```
jcox-dev: Interesting additional pathing for this endpoint. Could this be the 'PUT' for `/respondents/id/<id>` or simplified to `/respondents/id/<id>/update`?

test/test_respondent_controller.py
jcox-dev: @benjefferies I agree we should be asserting but the status check is already done in the helper function. We should probably be asserting the contents returned against expected instead.

```
data = self.change_respondent_details(respondent_id, payload, 200)
self.assertEqual(data, 'Successfully updated respondent details')
```

jcox-dev (changes requested):
Docstring amendment; endpoint path consideration; and comment regarding test of response text.

ras_party/views/respondent_view.py
necrophonic: Good point.

I wouldn't go as far as the extra `.../id/...` but `/respondents/<id>` would make sense under RESTful guidelines.

The "change" is implied through the method verb.

jwgwalton: +1 for  'PUT' for /respondents/id/<id>

jcox-dev (approved):

PR #110, opened 2018-03-13
Filter party associations by survey id

jwgwalton:


ras_party/controllers/party_controller.py
```diff
@@ -74,3 +74,30 @@ def get_party_by_id(sample_unit_type, id, session):
...
+            filtered_association['enrolments'] = filtered_enrolments
+            filtered_associations.append(filtered_association)
+
+    business['associations'] = filtered_associations
+
+    return business
+
+
+def filter_enrolments(existing_enrolments, survey_id):
+    filtered_enrolments = []
```
jcox-dev, line 99: 🎨 A list comprehension would be preferred here:

```python
return [enrolment for enrolment in existing_enrolments if enrolment['surveyId'] == survey_id]
```

ras_party/controllers/party_controller.py
```diff
@@ -74,3 +74,30 @@ def get_party_by_id(sample_unit_type, id, session):
...
+
+
+def get_business_with_respondents_filtered_by_survey(sample_unit_type, id, survey_id):
+    business = get_party_by_id(sample_unit_type, id)
+
+    filtered_associations = []
+    for association in business['associations']:
+
+        filtered_association = {'partyId': association['partyId']}
+
```
jcox-dev, line 86: :art: Can probably remove this extra line.

jcox-dev (approved):
Couple of comments but otherwise works as expected.

ras_party/controllers/party_controller.py
jwgwalton: I'm split on list comprehensions, sometime I think they make it less comprehendible. That one is pretty clear though. The end result of this babbling is that you're right and I will change it.

pricem14pc (approved):
Tested in PREPROD

PR #111, opened 2018-03-14
Add makefile and tidy code

insacuri:
Added a lint option that does both flake8 and checks for security vulnerabilities.
Updated readme with new ways to do things

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build lint start
+.PHONY: build lint start test
 
 build:
 	pipenv install --dev
--- a/Makefile
+++ b/Makefile
@@ -7,5 +7,8 @@ lint:
 	pipenv run flake8 ./ras_party ./test
 	pipenv check ./ras_party ./test
 
+test: lint
+	pipenv run tox
+
 start:
 	pipenv run python3 run.py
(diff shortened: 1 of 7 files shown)


Pipfile
```diff
@@ -28,6 +28,7 @@ retrying = "==1.3.3"
 sqlalchemy = "==1.1.10"
 sqlalchemy-utils = "==0.32.14"
 structlog = "==17.2.0"
+"flake8" = "*"
```
andrewmil: This should be in `dev-packages` not `packages`

Makefile
```diff
@@ -0,0 +1,11 @@
...
+
+build:
+	pipenv install --dev
+
+lint:
+	pipenv run flake8 ./ras_party ./test
+	pipenv check ./ras_party ./test
+
+start:
+	pipenv run python3 run.py
```
andrewmil: Should we have a command for running tests as well?

andrewmil (approved):
Looks good 👍 
Just left a couple small comments

Pipfile
insacuri: Removed, was already in dev-packages anyway

Makefile
insacuri: Added a test command

PR #112, opened 2018-03-16
Change email short term

rhiGriff:


ras_party/controllers/queries.py
```diff
@@ -80,10 +80,17 @@ def update_respondent_details(respondent_data, respondent_id, session):
 
     logger.debug('Updating respondent details', respondent_id=respondent_id)
 
-    session.query(Respondent).filter(Respondent.party_uuid == respondent_id).update({
-                                     Respondent.first_name: respondent_data['firstName'],
-                                     Respondent.last_name: respondent_data['lastName'],
-                                     Respondent.telephone: respondent_data['telephone']})
+    if str(Respondent.first_name) != str(respondent_data['firstName']) or str(Respondent.last_name) != \
```
Ogreman: Surely this will always evaluate to True as it's comparing to the class attributes (rather than an instance) of Respondent? Perhaps retrieve the instance (using the `party_uuid == respondent_id`) first then check for differences against that.

jcox-dev: Surely this will always evaluate to True as it's comparing to the class attributes (rather than an instance) of Respondent? Perhaps retrieve the instance (using the party_uuid == respondent_id) first then check for differences against that.

jcox-dev (changes requested):

test/test_respondent_controller.py
```diff
@@ -107,17 +107,33 @@ def test_update_respondent_details_success(self):
...
+            "telephone": "07837230942",
+            "email_address": "a@b.com",
+            "new_email_address": "john.snow@thisemail.com"
             }
         self.change_respondent_details(respondent_id, payload, 200)
 
     def test_update_respondent_details_respondent_does_not_exist_error(self):
-        self.populate_with_respondent()
-        respondent_id = ''
+        # self.populate_with_respondent()
```
gemmai95: Why is this comment here?

gemmai95 (approved):

LJBabbage (approved):
As agreed with tech lead this isn't ideal and needs a rewrite, but can be merged for now

tej99 (approved):

jcox-dev (approved):

rhiGriff:
[US058: INT Change Respondent Email Address [XL]](https://trello.com/c/KQUOeZBV/46-us058-int-change-respondent-email-address-xl)

PR #121, opened 2018-04-12
Us108 display trading as

jwgwalton:
Persist trading as to DB rather than concatenate trading styles in frontend.

https://github.com/ONSdigital/ras-backstage/pull/94

https://github.com/ONSdigital/response-operations-ui/pull/121

### DIFF
--- a/migrations/versions/dfeba9ccfc46_create_trading_as_field_in_business_.py
+++ b/migrations/versions/dfeba9ccfc46_create_trading_as_field_in_business_.py
@@ -1,0 +1,35 @@
+"""create trading_as field in business_attributes
+
+Revision ID: dfeba9ccfc46
+Revises: 2798b5d5566a
+Create Date: 2018-04-12 12:11:59.813999
+
+"""
+from alembic import op
+import json
+
+
+# revision identifiers, used by Alembic.
+revision = 'dfeba9ccfc46'
+down_revision = '2798b5d5566a'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+
+    sql_query = "SELECT * FROM partysvc.business_attributes"
+    result = conn.execute(sql_query)
+    for row in result:
+        attributes = json.loads(json.dumps(row['attributes']))
+        trading_as = '{tradstyle1} {tradstyle2} {tradstyle3}'.format(**attributes)
+        attributes['trading_as'] = ' '.join(trading_as.split())
+        json_attributes = json.dumps(attributes)
+        insert_sql = f"UPDATE partysvc.business_attributes " \
+                     f"SET attributes = '{json_attributes}' WHERE id={row['id']}"
+        conn.execute(insert_sql)
+
+
+def downgrade():
+    pass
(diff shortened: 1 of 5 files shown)


jwgwalton:
[US108: Display Trading As on RU Details](https://trello.com/c/KwObK7ly/181-us108-display-trading-as-on-ru-details)

migrations/versions/dfeba9ccfc46_create_trading_as_field_in_business_.py
```diff
@@ -0,0 +1,35 @@
...
+
+def upgrade():
+    conn = op.get_bind()
+
+    sql_query = "SELECT * FROM partysvc.business_attributes"
+    result = conn.execute(sql_query)
+    for row in result:
+        attributes = json.loads(json.dumps(row['attributes']))
+        trading_as = '{tradstyle1} {tradstyle2} {tradstyle3}'.format(**attributes)
+        attributes['trading_as'] = ' '.join(trading_as.split())
```
tej99, line 27: why are you doing a split and then join again here?  attributes['trading_as'] = trading_as?

ras_party/models/models.py
```diff
@@ -66,7 +66,9 @@ def from_party_dict(party):
         ba = BusinessAttributes(business_id=b.party_uuid, sample_summary_id=party['sampleSummaryId'])
         ba.attributes = party.get('attributes')
         name = '{runame1} {runame2} {runame3}'.format(**ba.attributes)
+        trading_as = '{tradstyle1} {tradstyle2} {tradstyle3}'.format(**ba.attributes)
         ba.attributes['name'] = ' '.join(name.split())
+        ba.attributes['trading_as'] = ' '.join(trading_as.split())
```
tej99, line 71: Am I missing something here?  ' '.join(trading_as.split()) is used everywhere trading_as is being assigned but surely it just returns the same thing :/

tej99 (changes requested):
LG just one question

migrations/versions/dfeba9ccfc46_create_trading_as_field_in_business_.py
jwgwalton: gets rid of any spaces if any of them are empty.

andrewmil: Better solution to this is `trading_as.strip()` removes leading and trailing spaces

tej99 (approved):

migrations/versions/dfeba9ccfc46_create_trading_as_field_in_business_.py
tej99: It's the double space inbetween tradstyle1 and tradstyle3 if tradstyle2 is empty. That's what I understood anyway

jwgwalton: with strip if tradstyle2 is blank it will end up with "tradstyle1 tradstyle2" rather than "tradstyle1tradstyle2"

andrewmil (approved):
👍 LGTM

PR #129, opened 2018-05-08
Change search for respondent by email to return created users

gemmai95:
For US115 this allows respondent to be searched for using email to return created respondents.


gemmai95:
[US115: View a respondents account details](https://trello.com/c/xI1qPlfC/83-us115-view-a-respondents-account-details)

tej99 (changes requested):
Will we want to search for a respondent without the whole email address and return all the emails similar to that string? If so might be better to just create a new query for searching respondents.

gemmai95:
@tej99 we will but not atm this is the first iteration. Looking into this now.

tej99 (approved):

emilioward (approved):

PR #130, opened 2018-05-11
Simplify party controller logic

insacuri:
The IsIn part of the validator seems needlessly complicated considering all that's being checked is that the sampleUnitType is either B or BI. 
This PR keeps the existing functionality while making it more readable and removing the need for extra classes.

EDIT:  It would be wise to have someone else run the acceptance tests against this just to make sure none of the functionality changed.

### DIFF
--- a/ras_party/controllers/party_controller.py
+++ b/ras_party/controllers/party_controller.py
@@ -2,7 +2,6 @@ from flask import current_app
 
 from ras_party.controllers.queries import query_business_by_party_uuid, query_business_by_ref
 from ras_party.controllers.queries import query_respondent_by_party_uuid
-from ras_party.controllers.validate import Validator, IsIn
 from ras_party.exceptions import RasError
 from ras_party.models.models import Business, Respondent
 from ras_party.support.session_decorator import with_db_session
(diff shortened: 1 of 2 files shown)


ras_party/controllers/party_controller.py
```diff
@@ -57,23 +54,20 @@ def get_party_by_ref(sample_unit_type, sample_unit_ref, session):
...
     elif sample_unit_type == Respondent.UNIT_TYPE:
         respondent = query_respondent_by_party_uuid(id, session)
         if not respondent:
             return RasError("Respondent with id does not exist.", respondent_id=id, status=404)
+    else:
+        raise RasError(f"{sample_unit_type} is not a valid value for sampleUnitType. Must be one of ['B', 'BI']",
+                       status=400)
 
-        return respondent.to_party_dict()
+    return respondent.to_party_dict()
```
andrewmil: Should we have this return statement be inside the `elif`? That makes it more symmetrical with the `business.to_party_dict` above it

andrewmil (approved):
LGTM all the acceptance tests still pass locally for me

tej99 (approved):

PR #131, opened 2018-05-14
Us053 remove loaded sample

tej99:
- Add ability to search business against trading as as well as the name
- Filter out 'inactive' attributes from search and returned business data

Inactive attributes are attributes that have no collection exercise associated to them. If a Business has no active attributes then the get api should return a 404 even if the business itself is in the database.

https://trello.com/c/IQ8agznp/86-us053-int-remove-a-loaded-sample-13


tej99:
[US053: INT- Remove a loaded sample (13)](https://trello.com/c/IQ8agznp/86-us053-int-remove-a-loaded-sample-13)

ras_party/models/models.py
```diff
@@ -139,7 +156,11 @@ def _get_attributes_for_collection_exercise(self, collection_exercise_id=None):
                 if attributes.collection_exercise == collection_exercise_id:
                     return attributes
 
-        return self.attributes[-1]
+        try:
+            return next((attributes for attributes in self.attributes if attributes.collection_exercise))
+        except StopIteration:
+            logger.error("No active attributes for business", party_id=self.party_uuid)
+            raise RasError("Business with reference does not exist.", reference=self.business_ref, status=404)
```
andrewmil: Is this error message correct? In this case the business exists in the system but is no longer linked to an exercise?

andrewmil (approved):
LGTM :+1:

ras_party/models/models.py
```diff
@@ -139,7 +156,12 @@ def _get_attributes_for_collection_exercise(self, collection_exercise_id=None):
                 if attributes.collection_exercise == collection_exercise_id:
                     return attributes
 
-        return self.attributes[-1]
+        try:
+            return next((attributes for attributes in self.attributes if attributes.collection_exercise))
+        except StopIteration:
+            logger.error("No active attributes for business", party_id=self.party_uuid)
```
rhiGriff, line 162: Can we add the status code to this log line please as this will help in the splunk logs.

emilioward (approved):

PR #133, opened 2018-05-17
DO NOT MERGE: fixing verification bug

rhiGriff:
This PR makes 2 small changes:

The first change removes the setting of a users account to verified in the django server when they change a password.

The second stops a change password email being sent to a user if their account is not in the ACTIVE state in the party database.


ras_party/controllers/account_controller.py
```diff
@@ -275,23 +274,24 @@ def request_password_change(payload, session):
...
-        'RESET_PASSWORD_URL': verification_url,
-        'FIRST_NAME': respondent.first_name
-    }
+        personalisation = {
+            'RESET_PASSWORD_URL': verification_url,
+            'FIRST_NAME': respondent.first_name
+        }
 
-    logger.info('Reset password url', url=verification_url)
+        logger.info('Reset password url', url=verification_url)
```
jcox-dev: Maybe we can log the party_id with the generated verification_url here?

jcox-dev (changes requested):
Does and checks what it says on the tin. 

Included a suggestion for linking the party with the notify URL in the logs.

rhiGriff:
[Receiving 500 error when trying to verify email address.](https://trello.com/c/Zpr07DyQ/221-receiving-500-error-when-trying-to-verify-email-address)

ras_party/controllers/account_controller.py
```diff
@@ -275,23 +274,24 @@ def request_password_change(payload, session):
     if not respondent:
         raise RasError("Respondent does not exist.", status=404)
 
-    email_address = respondent.email_address
-    verification_url = PublicWebsite().reset_password_url(email_address)
+    if respondent.status is RespondentStatus.ACTIVE:
```
gemmai95: I think we need some debug level logging on this. Just a "Requesting password change" and "Successfully changed password" with party_id potentially

test/test_respondent_controller.py
```diff
@@ -55,6 +55,24 @@ def populate_with_respondent(self, tran, session, respondent=None):
         account_controller.register_user(respondent, tran)
         return self.respondent
 
+    @transactional
+    @with_db_session
+    def populate_with_active_respondent(self, tran, session, respondent=None):
```
gemmai95: I don't know that a completely new function to add an active respondent is necessary just pass it in to the populate with respondent that's why it has the respondent.get('status'). This seems unnecessary when the same can be done with code already there.

ras_party/controllers/account_controller.py
```diff
@@ -275,23 +274,29 @@ def request_password_change(payload, session):
...
+        try:
+            NotifyGateway(current_app.config).request_password_change(
+                email_address, personalisation, str(party_id))
+        except RasNotifyError:
+            # Note: intentionally suppresses exception
+            logger.error('Error sending notification email for party_id', party_id=party_id)
+
+        logger.debug('Password reset email successfully sent', party_id=respondent.party_uuid)
 
     return {'response': "Ok"}
```
tej99, line 301: Should we still be returning a response ok when we haven't actually done what the call is expecting us to do (send a password change link)?

test/test_respondent_controller.py
jcox-dev: 👍 Could be a partial instead (if we updated the params for `populate_with_respondent`:
```python
...
    populate_with_active_respondent = functools.partialmethod(populate_with_respondent, status=RespondentStatus.ACTIVE)
...
```

ras_party/controllers/account_controller.py
rhiGriff: Yes this is as Dan has said it should carry on to the same screen etc just not send the email

test/test_respondent_controller.py
gemmai95: We have an active respondent but yeah we could.

jcox-dev (approved):

gemmai95 (approved):

tej99 (approved):

liamtoozer (approved):

PR #134, opened 2018-05-17
Add batch calls for respondents and businesses

warrenbailey:
Adding routes for respondents and business to allow clients to get details of multiple parties.

The two routes each take an ids request parameter which allows the caller to limit the result set to the required ids. This parameter is delimited by ',' and if not provided then this function returns the entire results set.

For example the rest call for respondents would be:
http://localhost:8081/party-api/v1/respondents?ids=e019ae8a-56ff-4bd7-ac58-8b7d09e9ecd1,6dd79b0f-0115-4887-979e-275ddfa078c9

While the call for business would be:
http://localhost:8081/party-api/v1/businesses?ids=350fc0ff-031f-4fde-9200-4eb5ab792b3d,4135017a-a533-4a16-8ed6-9a1ada4a71b4


warrenbailey:
Codacy error is because it's using the with_db_session function wrapper and that automatically injects the session object

insacuri:
For anyone looking at that codacy fail, that 'missing variable in function call' happens because of the way the '@with_db_session' injects the session and isn't really an issue.

insacuri:
LGTM, though I have a question about it.  Is there any downside to returning every respondant/business?  My gut says there might be, as reading all that in when it gets to be a certain size could crash the application.  I would've thought that if you don't provide any IDs then you'd return an error because it's probably a mistake

test/test_party_controller.py
```diff
@@ -38,6 +38,82 @@ def test_get_business_by_id_returns_correct_representation(self):
...
+            .attributes(source='test_get_business_by_ids_returns_correct_representation') \
+            .as_business()
+
+        mock_business_2 = MockBusiness() \
+            .attributes(source='test_get_business_by_ids_returns_correct_representation') \
+            .as_business()
+
+        party_id_1 = self.post_to_businesses(mock_business_1, 200)['id']
+        party_id_2 = self.post_to_businesses(mock_business_2, 200)['id']
+        self.assertNotEquals(party_id_1, party_id_2)
```
LJBabbage: This seems a bit of overkill

test/test_party_controller.py
```diff
@@ -38,6 +38,82 @@ def test_get_business_by_id_returns_correct_representation(self):
...
```
LJBabbage: As above

ajmaddaford:
Secure message implements multiple url parameters of the same name as `param=a&param=b` - https://github.com/ONSdigital/ras-secure-message/blob/0000cdcda87ac38d2aa4fefec9c0fdd56f17de3c/secure_message/resources/messages.py#L194 - it would be good to have a consistent approach for this.

test/test_party_controller.py
```diff
@@ -38,6 +38,82 @@ def test_get_business_by_id_returns_correct_representation(self):
...
+
+        self.assertEqual(response[0].get('id'), party_id_1)
+        self.assertEqual(response[0].get('sampleSummaryId'), mock_business_1['sampleSummaryId'])
+        self.assertEqual(response[0].get('name'), mock_business_1.get('name'))
+
+        self.assertEqual(response[1].get('id'), party_id_2)
+        self.assertEqual(response[1].get('sampleSummaryId'), mock_business_2['sampleSummaryId'])
+        self.assertEqual(response[1].get('name'), mock_business_2.get('name'))
+
+    def test_get_business_by_ids_with_unknown_id_returns_correct_representation(self):
```
LJBabbage: Is this right, unknown id returns 2 
self.assertEquals(len(response), 2)

LJBabbage (changes requested):

test/test_respondent_controller.py
```diff
@@ -103,6 +103,90 @@ def test_get_respondent_by_id_returns_correct_representation(self):
...
+        self.assertEquals(len(response), 1)
+        self.assertEqual(response[0]['emailAddress'], 'res1@example.com')
+
+    def test_get_respondent_by_ids_with_only_unknown_id_returns_none(self):
+        self.populate_with_respondent()
+        party_uuid = str(uuid.uuid4())
+        response = self.get_respondents_by_ids(party_uuid)
+        self.assertEquals(len(response), 0)
+
+    def test_get_respondent_by_ids_with_unknown_id_returns_correct_representation(self):
```
LJBabbage: Same as above, name wrong?

LJBabbage (changes requested):

test/test_respondent_controller.py
warrenbailey: As above

test/test_party_controller.py
warrenbailey: Yes because 2 out of the 3 ids are know. I'll change the name to make it more descriptive.

test/test_party_controller.py
warrenbailey: As above

test/test_party_controller.py
warrenbailey: Maybe, but if someone changes how the ids are generated and they suddenly become the same the rest of the test is pointless. I don't see the harm in having it.

LJBabbage (approved):

LJBabbage (dismissed):
Github gone nuts

LJBabbage (changes requested):
Failing tests

ras_party/controllers/queries.py
```diff
@@ -8,6 +8,25 @@
 logger = structlog.wrap_logger(logging.getLogger(__name__))
 
 
+def query_businesses(session):
```
insacuri: Not used anymore

ras_party/controllers/queries.py
```diff
@@ -30,6 +49,27 @@ def query_business_by_ref(business_ref, session):
     return session.query(Business).filter(Business.business_ref == business_ref).first()
 
 
+def query_respondents(session):
```
insacuri: Not used anymore

ras_party/controllers/respondent_controller.py
```diff
@@ -1,11 +1,37 @@
...
-    query_respondent_by_email, update_respondent_details
+    query_respondent_by_email, update_respondent_details, query_respondent_by_party_uuids, query_respondents
 from ras_party.controllers.validate import Validator, IsUuid
 from ras_party.exceptions import RasError
 from ras_party.support.session_decorator import with_db_session
 from ras_party.controllers.account_controller import change_respondent
 
 
+@with_db_session
+def get_respondents(session):
```
insacuri: Not used anymore

warrenbailey:
The merge to master has broken the tests. I'll have to adjust them.

LJBabbage (approved):

insacuri (approved):

ajmaddaford (changes requested):
When calling the respondents endpoint with an invalid uuid I get an unexpected error message - `There was an error committing the changes to the database.`

ajmaddaford (approved):

PR #135, opened 2018-05-18
fixing log line for splunk

rhiGriff:
Adding in a logger.exception log line to handle the exception temporarily so that the exception can be seen in Splunk before a larger refactor is done on the logging in the repo.


rhiGriff:
[Investigate database error in party not logging the exception.](https://trello.com/c/FjvOPDKL/232-investigate-database-error-in-party-not-logging-the-exception)

pricem14pc (approved):

tej99 (dismissed):

ras_party/support/session_decorator.py
```diff
@@ -32,6 +32,7 @@ def wrapper(*args, **kwargs):
             raise
         except Exception as e:
             logger.info("Rolling-back database session.")
```
benjefferies: I don't think we should be logging at info level if we raise an error at all, it's misleading. Lets remove this line and incorporate it into the exception logging

benjefferies (dismissed):

ras_party/support/session_decorator.py
tej99: I agree but Rhi mentioned @davidmort wanted both in?

davidmort: not quite - what i wanted was the error to be logged out as it's just currently returned in the api response https://github.com/ONSdigital/ras-party/blob/f5d645226ba4803d9509263517746208e4b0e731/ras_party/error_handlers.py. adding the two log lines into one as suggested sounds good, but really we should fix the error_handlers.py to log out errors/exceptions correctly

tej99: Yeah for that we'll need to remove that generic RasError

benjefferies: Spoken to @rhiGriff and she's got some other cards she's working on so I can help out with this

ras_party/support/session_decorator.py
```diff
@@ -1,15 +1,35 @@
...
+def handle_session(f, args, kwargs):
+    logger.info("Acquiring database session.")
+    session = current_app.db.session()
+    try:
+        result = f(*args, session=session, **kwargs)
+        logger.info("Committing database session.")
+        session.commit()
+        return result
+    except RasError:
+        logger.error(f"Rolling back database session due to failure executing function", function=f.__name__)
```
gemmai95: I don't see the point of logging at error here then exception in the error handler.

ras_party/support/session_decorator.py
```diff
@@ -1,15 +1,35 @@
...
+        session.rollback()
+        raise
+    except Exception as e:
+        logger.error("Rolling back database session due to uncaught exception", exception_type=e.__name__)
```
gemmai95: So if we log this specific error message here why do we need a generic one in the exception handler? This seems like a better log line as it's more informative of what is actually going on.

ras_party/support/session_decorator.py
benjefferies: Fair point the function info will be in the traceback

benjefferies: Actually rethinking. It does have the context of rolling back the session so I think it's worth being here but not the bit about the function

ras_party/support/session_decorator.py
benjefferies: I think this scenario isn't the norm and it's hard to justify in relation to it. The normal case we'd catch and exception and handle it appropriately whether that's showing an error page or returning a bad status but it's hard to get that context out of the decorator. The other thing we could do is catch the exception in all of the views but that seems ugly to me

gemmai95 (changes requested):
Travis is complaining of a lot of failing tests these will need addressed.

ras_party/error_handlers.py
```diff
@@ -12,24 +12,24 @@
...
 
-@blueprint.app_errorhandler(HTTPError)
+@blueprint.app_errorhandler(RequestException)
 def http_error(error):
-    detail = error.response.json().get('detail', "No further detail.")
-    response = jsonify({'errors': [str(error), detail]})
+    errors = {'errors': {'method': error.request.method, 'url': error.request.url, }}
+    response = jsonify(errors)
     response.status_code = 500
+    logger.exception('Uncaught exception', errors=errors, status=500)
```
JamesGardiner: This will also catch 4xx errors - shouldn't we get the status from the error response?

ras_party/support/session_decorator.py
JamesGardiner: What if the error is related to a failure between the engine and the database server? We probably need better error handling around those types of exceptions so that we are able to retry in the correct circumstances (i.e. TCP connection unexpectedly closed, network issues)

JamesGardiner (changes requested):

ras_party/error_handlers.py
benjefferies: Oops I meant to do that 🤔

ras_party/support/session_decorator.py
benjefferies: This should probably be handled at the source of the exception. I was thinking some more any maybe we should just get rid of this session_decorator.py and create transactions and handle them using http://flask.pocoo.org/docs/0.12/api/#flask.Flask.before_request and http://flask.pocoo.org/docs/0.12/api/#flask.Flask.after_request

benjefferies:
Tests are now passing @gemmai95. I've fixed the points you mentioned @JamesGardiner

gemmai95 (approved):

JamesGardiner (approved):

PR #137, opened 2018-05-24
Remove exists in validator class

insacuri:
The validator class is pretty needless, as it's possible to generate the same errors in code, rather then pass it to a class and have it figure out if everything is okay or not.

This PR aims to preserve the current functionality of the exists validation.  v.errors produces a list of errors for RasError, though RasError will convert a string into a list of 1 error if given a string; Again, preserving the current functionality.

The unit tests pass but I've not yet been able to run this against the acceptance test suite, though a review of the code would be useful.


MIslam92 (changes requested):
This looks good to me just one thing unrelated to this PR though out some the files changed there is a lot of missing docstrings and unused parameters being passed in i.e. account_controller session etc. 

It would be worth doing a quick clean up of this now rather then never getting around to doing it?

MIslam92 (approved):
Disregard above created separate card for it otherwise
LGTM

ryangrundy7 (approved):
Looks good! 👍

PR #138, opened 2018-05-24
Remove isUuid from validator class and code.

insacuri:
This PR strips more functionality from the validator class in favour of generating the error directly in the code.  This should make reading the code a easier.

Unit tests pass but acceptance tests haven't been run against it.  Early review on this would be useful however.


MIslam92 (approved):
LGTM

benjefferies (approved):
Looks good and doesn't cause any regression

benjefferies:
@insacuri I've ran the acceptance tests against this branch

PR #141, opened 2018-06-07
Make db connection pool configurable

tej99:
To improve performance of the system the db connection pool needs to be increased to sustain the load. This PR allows us to configure the pool size with environment variables.

### DIFF
--- a/config.py
+++ b/config.py
@@ -125,9 +125,6 @@ class TestingConfig(DevelopmentConfig):
     REQUESTS_GET_TIMEOUT = 99
     REQUESTS_POST_TIMEOUT = 99
     DATABASE_SCHEMA = 'partysvc'
-    DB_POOL_SIZE = 5
-    DB_MAX_OVERFLOW = 10
-    DB_POOL_RECYCLE = -1
 
     RAS_PUBLIC_WEBSITE_URL = 'http://dummy.ons.gov.uk'
     RAS_CASE_SERVICE = 'http://mockhost:1111'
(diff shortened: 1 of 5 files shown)


tej99:
[Optimise DB Connection Settings](https://trello.com/c/SwXkyTRo/44-optimise-db-connection-settings)

test/test_create_database.py
```diff
@@ -16,7 +16,7 @@ def test_postgres_database_schema_exists(self):
         db_schema = 'partysvc'
 
         with patch('run.create_engine'), patch('run.scoped_session'), patch('run.command.upgrade') as upgrade:
-            create_database(db_connection, db_schema)
+            create_database(db_connection, db_schema, 5, 10, -1)
```
gemmai95, line 19: Are these numbers meant to be configurable by the env vars?

tej99: Keeping them hard coded here is fine as the tests focus on the alembic stuff.

config.py
```diff
@@ -121,6 +125,9 @@ class TestingConfig(DevelopmentConfig):
     REQUESTS_GET_TIMEOUT = 99
     REQUESTS_POST_TIMEOUT = 99
     DATABASE_SCHEMA = 'partysvc'
+    DB_POOL_SIZE = 5
+    DB_MAX_OVERFLOW = 10
+    DB_POOL_RECYCLE = -1
```
andrewmil: We don't need these in the Testing config as they'll use the default values in the Config class

andrewmil (approved):
:+1: Looks good

gemmai95 (approved):

PR #148, opened 2018-07-05
Fix resend verification email

gavinleeedwards:
# Motivation and Context
Service desk call TASK0241375

A bug exists where if a respondent has requested a change of email address then requested the verification email to be resent, the verification email was being sent to the user's old email address and the link did not work.

# How to test?
- Run alongside branch `fix-resend-verification-email-bug` in response-operations (https://github.com/ONSdigital/response-operations-ui/pull/203)

- Amend a current respondent's email address
- Verify the email by using the link in the party service logs
- The account should be verified:
The old email address should no longer work when logging in through frontstage
The respondent should be able to log in using the new email address
The email address should be updated against the respondent's details in response-ops

# Links
Trello card: https://trello.com/c/p6cE41dU
Related PR: https://github.com/ONSdigital/response-operations-ui/pull/203

### DIFF
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -479,7 +479,7 @@ def _send_email_verification(party_id, email):
     Send an email verification to the respondent
     """
     verification_url = PublicWebsite().activate_account_url(email)
-    logger.info('Verification URL for party_id', party_id=party_id, url=verification_url, email=email)
+    logger.info('Verification URL for party_id', party_id=party_id, url=verification_url)
 
     personalisation = {
         'ACCOUNT_VERIFICATION_URL': verification_url
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -410,7 +410,10 @@ def resend_verification_email(party_uuid, session):
     if not respondent:
         raise RasError(NO_RESPONDENT_FOR_PARTY_ID, status=404)
 
-    _send_email_verification(party_uuid, respondent.email_address)
+    if respondent.pending_email_address != '':
+        _send_email_verification(party_uuid, respondent.pending_email_address)
+    else:
+        _send_email_verification(party_uuid, respondent.email_address)
 
     return {'message': EMAIL_VERIFICATION_SENT}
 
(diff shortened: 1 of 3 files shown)


ras_party/controllers/account_controller.py
```diff
@@ -410,7 +410,10 @@ def resend_verification_email(party_uuid, session):
     if not respondent:
         raise RasError(NO_RESPONDENT_FOR_PARTY_ID, status=404)
 
-    _send_email_verification(party_uuid, respondent.email_address)
+    if respondent.pending_email_address != '':
```
shahi645: Probably better as `if not respondent.pending_email_address` So it deals with `Null` values too

ras_party/controllers/account_controller.py
```diff
@@ -410,7 +410,10 @@ def resend_verification_email(party_uuid, session):
...
+        _send_email_verification(party_uuid, respondent.pending_email_address)
+    else:
+        _send_email_verification(party_uuid, respondent.email_address)
 
     return {'message': EMAIL_VERIFICATION_SENT}
```
shahi645, line 418: Is there any exception where the verification email would not be sent, if so worth putting this  if statement in a `try/except` block. Otherwise if the exception occurs, we are still gonna get the `EMAIL_VERIFICATION_SENT` message

shahi645 (dismissed):
Looks good @gavinleeedwards Just left a couple minor things for consideration

shahi645 (changes requested):
Comments left for requested changes

ras_party/controllers/account_controller.py
```diff
@@ -410,7 +410,10 @@ def resend_verification_email(party_uuid, session):
     if not respondent:
         raise RasError(NO_RESPONDENT_FOR_PARTY_ID, status=404)
 
-    _send_email_verification(party_uuid, respondent.email_address)
+    if respondent.pending_email_address != '':
+        _send_email_verification(party_uuid, respondent.pending_email_address)
+    else:
+        _send_email_verification(party_uuid, respondent.email_address)
```
gemmai95: This doesn't seem to be being tested.

ras_party/controllers/account_controller.py
gavinleeedwards: If there's no pending email then it comes through as an empty string rather than null so doing `if not respondent.pending_email_address` wouldn't work.

shahi645: `if not` will pick up on empty string, but yeah if its just an empty string can be left like this i guess

gavinleeedwards: Done 👍

ras_party/controllers/account_controller.py
gavinleeedwards: The only exception gets handled in _send_email_verification.  I've kept it in line with similar calls.

shahi645 (approved):
Tested and works well, still one test to add but happy to approve functionality

gemmai95 (approved):

PR #149, opened 2018-07-24
Remove bi/send no active enrolment event

andrewmil:
# Motivation and Context
When no active enrolments are remaining for a business/survey we need to updated cases in the case service. Currently we send a `DISABLE_RESPONDENT_ENROLMENT` case event against each `BI` case we are disabling. As we are moving toward the removal of `BI` cases we need an equivalent which will work for `B` cases.

# What has changed
A new `NO_ACTIVE_ENROLMENTS` case event is sent to case service when the final respondent is disabled for a business/survey. 
We continue sending `DISABLE_RESPONDENT_ENROLMENT` case events against the associated `BI` cases as these have not yet been removed from the system

# How to test?
Disable the final respondent for a business/survey either through the UI or the party service api, confirm that a case event with category `NO_ACTIVE_ENROLMENTS` is posted to the case service
See associated [case-service PR]() for testing instructions with all associated PR's

Unit tests should pass
Acceptance tests should pass when run alongside associated PR's

# Links
[Trello](https://trello.com/c/RlGDFpX8/18-disable-enrolment-stop-creating-b-cases-case-events-l)
### Associated PR's
[case-service](https://github.com/ONSdigital/rm-case-service/pull/59)
[casesvc-api](https://github.com/ONSdigital/rm-casesvc-api/pull/25)

!!! This shouldn't be deployed into production until the appropriate data migration has been done to ensure that one B case exists for every casegroup !!!


ras_party/controllers/queries.py
```diff
@@ -3,7 +3,7 @@
 from sqlalchemy import func, and_, or_
 import structlog
 
-from ras_party.models.models import Business, BusinessRespondent, Enrolment, Respondent, BusinessAttributes
+from ras_party.models.models import Business, BusinessRespondent, Enrolment, EnrolmentStatus, Respondent, BusinessAttributes
```
shahi645: @andrewmil apparently this line is too long 🤷‍♂️

shahi645:
Tested fine for me, waiting for @andrewmil to add tests

config.py
```diff
@@ -37,7 +37,7 @@ class Config(object):
         DATABASE_URI = cf.db.credentials['uri']
     else:
         DATABASE_SCHEMA = os.getenv('DATABASE_SCHEMA', 'partysvc')
-        DATABASE_URI = os.getenv('DATABASE_URI', "postgres://postgres:postgres@localhost:6432/postgres")
+        DATABASE_URI = os.getenv('DATABASE_URI', "postgres://postgres:postgres@localhost:5432/postgres")
```
shahi645: was this meant to be committed?

shahi645: ignore

shahi645 (approved):
LGTM

ras_party/controllers/account_controller.py
```diff
@@ -168,6 +169,16 @@ def change_respondent_enrolment_status(payload, session):
     for case in get_cases_for_collection_exercise(survey_id, business_id, respondent_id):
         post_case_event(case['id'], business_id, category=category, desc=description)
 
+    session.commit()
+    enrolment = query_enrolment_by_survey_business(business_id, survey_id, session)
+    if not enrolment:
+        logger.info('No active enrolments', business_id=business_id, survey_id=survey_id)
+        for case in get_cases_for_collection_exercise(survey_id, business_id, business_id):
+            post_case_event(case['id'],
+                            party_id=None,
```
gemmai95: Should we remove party_id here at some point as it will become redundant without BI plus you already have the case id?

andrewmil: You're right when we remove BI cases we'll no longer have to create new cases off the back of old ones and that'll mean we never have to send `party_id` as part of the case event request. Until we've made that change in case service I think we should leave it in here though. 
We still need to send a `party_id` for the `RESPONDENT_ENROLED` category for now

ras_party/controllers/account_controller.py
```diff
@@ -168,6 +169,16 @@ def change_respondent_enrolment_status(payload, session):
     for case in get_cases_for_collection_exercise(survey_id, business_id, respondent_id):
         post_case_event(case['id'], business_id, category=category, desc=description)
 
+    session.commit()
+    enrolment = query_enrolment_by_survey_business(business_id, survey_id, session)
+    if not enrolment:
+        logger.info('No active enrolments', business_id=business_id, survey_id=survey_id)
+        for case in get_cases_for_collection_exercise(survey_id, business_id, business_id):
```
ajmaddaford: - `business_id` is repeated
- `get_cases_for_collection_exercise` repeats the call already made on line 169

ras_party/controllers/account_controller.py
```diff
@@ -168,6 +169,16 @@ def change_respondent_enrolment_status(payload, session):
...
+            post_case_event(case['id'],
```
ajmaddaford: What happens if this works for some cases and then fails?

ras_party/controllers/queries.py
```diff
@@ -169,3 +170,17 @@ def query_enrolment_by_survey_business_respondent(respondent_id, business_id, su
...
+    """
+    Query to return enrolments based on business id and survey
+    :param business_id,
+    :param survey_id
+    :return: enrolment for survey and business for respondent
+    """
+    logger.debug('Querying enrolment', business_id=business_id, survey_id=survey_id)
+    response = session.query(Enrolment).filter(and_(Enrolment.business_id == business_id,
+                                                    Enrolment.survey_id == survey_id,
+                                                    Enrolment.status == EnrolmentStatus.ENABLED)).first()
```
ajmaddaford: If we're only concerned with if any enrolments exist, should this be a count, rather than the first record?

test/mocks.py
```diff
@@ -36,6 +36,8 @@ def __call__(self, uri, *args, **kwargs):
                         MockResponse(get_case_by_iac.response),
                     'http://mockhost:1111/cases/partyid/438df969-7c9c-4cd4-a89b-ac88cf0bfdf3':
                         MockResponse(get_cases_by_party.response),
+                    'http://mockhost:1111/cases/partyid/3b136c4b-7a14-4904-9e01-13364dd7b972':
+                        MockResponse(get_cases_by_party.response),
```
ajmaddaford, line 40: How does this test the new functionality?

ajmaddaford (changes requested):

ras_party/controllers/queries.py
andrewmil: :+1: yeah that makes it clearer what we're actually doing here

ras_party/controllers/account_controller.py
andrewmil: Bad stuff probably
We have a problem generally across the system where we have multiple api calls sequentially, if one fails in the middle of things we have don't have ways to rollback things that have been done in other services

ras_party/controllers/account_controller.py
andrewmil: So we have to do both of those things because there are two sets of cases we are returning (B's and BI's)
`get_cases_for_collection_exercise` is a nasty function which makes a bunch of calls because of how distributed our data is. I'm going to have a look at it to see if we can simplify

andrewmil: This stuff will be simplified significantly when we remove `BI` cases

ras_party/controllers/account_controller.py
```diff
@@ -155,19 +156,31 @@ def change_respondent_enrolment_status(payload, session):
...
 
+    casegroup_ids = get_business_survey_casegroups(survey_id, business_id)
+
     category = 'DISABLE_RESPONDENT_ENROLMENT' if change_flag == 'DISABLED' else 'ENABLE_RESPONDENT_ENROLMENT'
     description = "Disable respondent enrolment" if change_flag == 'DISABLED' else 'Enable respondent enrolment'
-    for case in get_cases_for_collection_exercise(survey_id, business_id, respondent_id):
+    for case in get_cases_for_casegroups(casegroup_ids, respondent_id):
         post_case_event(case['id'], business_id, category=category, desc=description)
 
+    session.commit()
```
ajmaddaford: What is this `session.commit()` for?

ras_party/controllers/queries.py
```diff
@@ -169,3 +170,17 @@ def query_enrolment_by_survey_business_respondent(respondent_id, business_id, su
                                                     Enrolment.business_id == business_id,
                                                     Enrolment.survey_id == survey_id)).first()
     return response
+
+
+def query_enrolment_by_survey_business(business_id, survey_id, session):
+    """
+    Query to return enrolments based on business id and survey
```
ajmaddaford: I think the method name should reflect it's a count, maybe `count_enrolments_by_survey_business`. Also, the docstring needs updating to reflect it's a count.

ajmaddaford (changes requested):

ras_party/controllers/account_controller.py
andrewmil: Above here we've disabled an enrolment with `enrolment.status = change_flag`
Then on the line below this one we query the database to get the number of enabled enrolments. If we don't commit first that previous enrolment wont be disabled yet

andrewmil: We could not commit and do a check for it there if the enrolment count was equal to '1' instead but I think it's better to actually disable the enrolment first

ajmaddaford: The commit is fine, it just wasn't clear what it was committing. It's much clearer now the commit has been moved next to the database change.

ajmaddaford (approved):
The changes make sense for where we're heading, accepting that there's further work to do to simplify things. 

I've raised a card to consider adding the survey id to the case, as that would remove a bunch of calls to other services to resolve the survey id.

PR #150, opened 2018-07-25
Fix logging levels

gavinleeedwards:
# Motivation and Context
There are a lot of errors/exceptions being reported in Splunk that aren't actual errors (e.g. someone typing in an enrolment code wrong isn't a problem with the application).  This work changes the logging levels to 'info' for exceptions that have no negative impact.

**This work only covers the logging levels - there's another piece of work to improve and  standardise the way that exceptions are raised and errors are logged across all services but a card has been raised to cover it. **

# What has changed
- Made some error messages more explicit to help when diagnosing errors in Splunk.
- Changed log levels to 'info' where appropriate
- Set up a new ClientError exception to log some exceptions at 'info' level
- Added new tests to cover the ClientError exception

# How to test?
- Tests should pass
- Generate a few exceptions.  They should be logging at 'info' level where appropriate.

### DIFF
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -89,8 +89,8 @@ def post_respondent(party, tran, session):
     business = query_business_by_party_uuid(business_id, session)
     if not business:
         raise ClientError("Could not locate business when creating business association",
-                       business_id=business_id,
-                       status=404)
+                          business_id=business_id,
+                          status=404)
 
     # Chain of enrolment processes
     translated_party = {
@@ -159,7 +159,7 @@ def change_respondent_enrolment_status(payload, session):
     respondent = query_respondent_by_party_uuid(respondent_id, session)
     if not respondent:
         raise ClientError("Respondent does not exist.  Unable to change enrolment status",
-                       respondent_id=respondent_id, status=404)
+                          respondent_id=respondent_id, status=404)
 
     enrolment = query_enrolment_by_survey_business_respondent(respondent_id=respondent.id,
                                                               business_id=business_id,
@@ -241,7 +241,7 @@ def change_respondent_password(token, payload, tran, session):
     respondent = query_respondent_by_email(email_address, session)
     if not respondent:
         raise ClientError("Respondent does not exist.",
-                       status=404)
+                          status=404)
 
     new_password = payload['new_password']
 
@@ -317,7 +317,7 @@ def change_respondent_account_status(payload, party_id, session):
     respondent = query_respondent_by_party_uuid(party_id, session)
     if not respondent:
         raise ClientError("Respondent does not exist.  Unable to change account status.", respondent_id=party_id,
-                       status=404)
+                          status=404)
     respondent.status = status
 
 
@@ -462,8 +462,8 @@ def add_new_survey_for_respondent(payload, tran, session):
         business = query_business_by_party_uuid(business_id, session)
         if not business:
             raise ClientError("Could not locate business when creating business association.",
-                           business_id=business_id,
-                           status=404)
+                              business_id=business_id,
+                              status=404)
         br = BusinessRespondent(business=business, respondent=respondent)
 
     enrolment = Enrolment(business_respondent=br,
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -301,7 +301,7 @@ def request_password_change(payload, session):
                 email_address, personalisation, str(party_id))
         except RasNotifyError:
             # Note: intentionally suppresses exception
-            logger.error('Error sending reset password email for party_id', party_id=party_id)
+            logger.error('Error sending reset password email for respondent_id', respondent_id=party_id)
 
         logger.debug('Password reset email successfully sent', party_id=respondent.party_uuid)
 
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -54,7 +54,7 @@ def post_respondent(party, tran, session):
     if 'id' in party:
         # Note: there's not strictly a requirement to be able to pass in a UUID, this is currently supported to
         # aid with testing.
-        logger.debug("'id' in respondent post message.")
+        logger.debug("'id' in respondent post message")
         try:
             uuid.UUID(party['id'])
         except ValueError:
@@ -65,11 +65,11 @@ def post_respondent(party, tran, session):
 
     iac = request_iac(party['enrolmentCode'])
     if not iac.get('active'):
-        raise ClientError("Enrolment code is not active.", status=400)
+        raise ClientError("Enrolment code is not active", status=400)
 
     existing = query_respondent_by_email(party['emailAddress'], session)
     if existing:
-        raise ClientError("Email address already exists.",
+        raise ClientError("Email address already exists",
                           status=400,
                           party_uuid=str(existing.party_uuid))
 
@@ -158,7 +158,7 @@ def change_respondent_enrolment_status(payload, session):
                 status=change_flag)
     respondent = query_respondent_by_party_uuid(respondent_id, session)
     if not respondent:
-        raise ClientError("Respondent does not exist.  Unable to change enrolment status",
+        raise ClientError("Respondent does not exist",
                           respondent_id=respondent_id, status=404)
 
     enrolment = query_enrolment_by_survey_business_respondent(respondent_id=respondent.id,
@@ -188,14 +188,14 @@ def change_respondent(payload, session):
     respondent = query_respondent_by_email(email_address, session)
 
     if not respondent:
-        raise ClientError("Respondent does not exist.", status=404)
+        raise ClientError("Respondent does not exist", status=404)
 
     if new_email_address == email_address:
         return respondent.to_respondent_dict()
 
     respondent_with_new_email = query_respondent_by_email(new_email_address, session)
     if respondent_with_new_email:
-        raise ClientError("New email address already taken.", status=409)
+        raise ClientError("New email address already taken", status=409)
 
     respondent.pending_email_address = new_email_address
 
@@ -218,7 +218,7 @@ def verify_token(token, session):
 
     respondent = query_respondent_by_email(email_address, session)
     if not respondent:
-        raise ClientError("Respondent does not exist.", status=404)
+        raise ClientError("Respondent does not exist", status=404)
 
     return {'response': "Ok"}
 
@@ -240,7 +240,7 @@ def change_respondent_password(token, payload, tran, session):
 
     respondent = query_respondent_by_email(email_address, session)
     if not respondent:
-        raise ClientError("Respondent does not exist.",
+        raise ClientError("Respondent does not exist",
                           status=404)
 
     new_password = payload['new_password']
@@ -250,7 +250,7 @@ def change_respondent_password(token, payload, tran, session):
         password=new_password)
 
     if oauth_response.status_code != 201:
-        raise RasError("Failed to change respondent password.")
+        raise RasError("Failed to change respondent password")
 
     personalisation = {
         'FIRST_NAME': respondent.first_name
@@ -265,7 +265,7 @@ def change_respondent_password(token, payload, tran, session):
         logger.error('Error sending notification email', respondent_id=party_id)
 
     # This ensures the log message is only written once the DB transaction is committed
-    tran.on_success(lambda: logger.info('Respondent has changed their password', party_id=party_id))
+    tran.on_success(lambda: logger.info('Respondent has changed their password', respondent_id=party_id))
 
     return {'response': "Ok"}
 
@@ -280,7 +280,7 @@ def request_password_change(payload, session):
 
     respondent = query_respondent_by_email(email_address, session)
     if not respondent:
-        raise ClientError("Respondent does not exist.", status=404)
+        raise ClientError("Respondent does not exist", status=404)
 
     logger.debug("Requesting password change", party_id=respondent.party_uuid)
 
@@ -302,7 +302,7 @@ def request_password_change(payload, session):
                 email_address, personalisation, str(party_id))
         except RasNotifyError:
             # Note: intentionally suppresses exception
-            logger.error('Error sending request to Notify Gateway for respondent_id', respondent_id=party_id)
+            logger.error('Error sending request to Notify Gateway', respondent_id=party_id)
 
         logger.debug('Password reset email successfully sent', party_id=respondent.party_uuid)
 
@@ -316,7 +316,7 @@ def change_respondent_account_status(payload, party_id, session):
 
     respondent = query_respondent_by_party_uuid(party_id, session)
     if not respondent:
-        raise ClientError("Respondent does not exist.  Unable to change account status.", respondent_id=party_id,
+        raise ClientError("Respondent does not exist", respondent_id=party_id,
                           status=404)
     respondent.status = status
 
@@ -393,7 +393,7 @@ def update_verified_email_address(respondent, tran, session):
         if rollback_response.status_code != 201:
             logger.error("Failed to rollback change to respondent email. Please investigate.",
                          party_id=respondent.party_uuid)
-            raise RasError("Failed to rollback change to respondent email.")
+            raise RasError("Failed to rollback change to respondent email")
 
     tran.compensate(compensate_oauth_change)
 
@@ -439,7 +439,7 @@ def add_new_survey_for_respondent(payload, tran, session):
 
     iac = request_iac(enrolment_code)
     if not iac.get('active'):
-        raise ClientError("Enrolment code is not active.", status=400)
+        raise ClientError("Enrolment code is not active", status=400)
 
     respondent = query_respondent_by_party_uuid(respondent_party_id, session)
 
@@ -461,7 +461,7 @@ def add_new_survey_for_respondent(payload, tran, session):
         """
         business = query_business_by_party_uuid(business_id, session)
         if not business:
-            raise ClientError("Could not locate business when creating business association.",
+            raise ClientError("Could not locate business when creating business association",
                               business_id=business_id,
                               status=404)
         br = BusinessRespondent(business=business, respondent=respondent)
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -89,7 +89,7 @@ def post_respondent(party, tran, session):
     business = query_business_by_party_uuid(business_id, session)
     if not business:
         raise ClientError("Could not locate business when creating business association",
-                          business_id=str(business_id),
+                          business_id=business_id,
                           status=404)
 
     # Chain of enrolment processes
@@ -159,7 +159,7 @@ def change_respondent_enrolment_status(payload, session):
     respondent = query_respondent_by_party_uuid(respondent_id, session)
     if not respondent:
         raise ClientError("Respondent does not exist",
-                          respondent_id=str(respondent_id), status=404)
+                          respondent_id=respondent_id, status=404)
 
     enrolment = query_enrolment_by_survey_business_respondent(respondent_id=respondent.id,
                                                               business_id=business_id,
@@ -300,7 +300,7 @@ def request_password_change(payload, session):
 
         try:
             NotifyGateway(current_app.config).request_password_change(
-                email_address, personalisation, str(party_id))
+                email_address, personalisation, party_id)
         except RasNotifyError:
             # Note: intentionally suppresses exception
             logger.error('Error sending request to Notify Gateway', respondent_id=party_id)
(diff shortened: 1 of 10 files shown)


ras_party/controllers/account_controller.py
```diff
@@ -65,11 +65,11 @@ def post_respondent(party, tran, session):
 
     iac = request_iac(party['enrolmentCode'])
     if not iac.get('active'):
-        raise RasError("Enrolment code is not active.", status=400)
+        raise ClientError("Enrolment code is not active.", status=400)
```
shahi645: 👍

ras_party/controllers/account_controller.py
```diff
@@ -230,13 +233,14 @@ def change_respondent_password(token, payload, tran, session):
...
     except SignatureExpired:
-        raise RasError('Expired email verification token', status=409, token=token)
+        raise ClientError('Expired email verification token', status=409, token=token)
     except (BadSignature, BadData) as e:
         raise RasError('Unknown email verification token', status=404, token=token, error=e)
 
     respondent = query_respondent_by_email(email_address, session)
     if not respondent:
-        raise RasError("Respondent does not exist.", status=404)
+        raise RasError("Respondent does not exist. Unable to change respondent's password",
```
shahi645: @gavinleeedwards Do we want info level here?

shahi645: ignore, just seen the `ClientError`

ras_party/error_handlers.py
```diff
@@ -5,13 +5,21 @@
...
-from ras_party.exceptions import RasError
+from ras_party.exceptions import ClientError, RasError
 
 logger = structlog.wrap_logger(logging.getLogger(__name__))
 
 blueprint = flask.Blueprint('error_handlers', __name__)
 
 
+@blueprint.app_errorhandler(ClientError)
+def client_error(error):
```
shahi645, line 16: 👍

ras_party/controllers/account_controller.py
```diff
@@ -297,7 +301,7 @@ def request_password_change(payload, session):
                 email_address, personalisation, str(party_id))
         except RasNotifyError:
             # Note: intentionally suppresses exception
-            logger.error('Error sending notification email for party_id', party_id=party_id)
+            logger.error('Error sending reset password email for party_id', party_id=party_id)
```
andrewmil: For consistency/clarity might be better to use `respondent_id=party_id`

ras_party/controllers/account_controller.py
```diff
@@ -297,7 +301,7 @@ def request_password_change(payload, session):
                 email_address, personalisation, str(party_id))
         except RasNotifyError:
             # Note: intentionally suppresses exception
-            logger.error('Error sending notification email for party_id', party_id=party_id)
```
gemmai95: Does this party_id need changed to respondent_id?

ras_party/support/session_decorator.py
```diff
@@ -21,12 +21,16 @@ def handle_session(f, args, kwargs):
...
+        session.rollback()
+        raise
     except RasError:
-        logger.error(f"Rolling back database session due to failure executing function")
+        logger.info(f"Rolling back database session due to failure executing function")
         session.rollback()
         raise
     except Exception:
-        logger.error("Rolling back database session due to uncaught exception")
+        logger.info("Rolling back database session due to uncaught exception")
```
andrewmil: We hit these quite a lot at the moment don't we? Would it be better for these to be at `debug` level?

andrewmil (commented):
Looks good left a couple of very small comments but it's might not be for this PR :+1:
Just running against acceptance tests now then I'll approve

andrewmil (approved):

ras_party/controllers/account_controller.py
gavinleeedwards: Changed the log message to respondent_id to make it clearer 👍

ras_party/controllers/account_controller.py
ajmaddaford: The error messages passed to `ClientError` aren't used - they don't appear in the logs.

ras_party/controllers/account_controller.py
```diff
@@ -65,11 +65,11 @@ def post_respondent(party, tran, session):
 
     iac = request_iac(party['enrolmentCode'])
     if not iac.get('active'):
-        raise RasError("Enrolment code is not active.", status=400)
+        raise ClientError("Enrolment code is not active.", status=400)
 
     existing = query_respondent_by_email(party['emailAddress'], session)
     if existing:
-        raise RasError("Email address already exists.", status=400, party_uuid=party.get('party_uuid', None))
+        raise ClientError("Email address already exists.", status=400, party_uuid=party.get('party_uuid', None))
```
ajmaddaford: Although `ClientError` persists the kwargs in a property, these aren't used when the exception is caught and the log line printed i.e. extra arguments like `party_uuid` won't be output in the logs.

ras_party/controllers/account_controller.py
```diff
@@ -208,13 +211,13 @@ def verify_token(token, session):
...
     except SignatureExpired:
-        raise RasError('Expired email verification token', status=409, token=token)
+        raise ClientError('Expired email verification token', status=409, token=token)
     except (BadSignature, BadData) as e:
         raise RasError('Unknown email verification token', status=404, token=token, error=e)
 
     respondent = query_respondent_by_email(email_address, session)
     if not respondent:
-        raise RasError("Respondent does not exist.", status=404)
+        raise RasError("Respondent does not exist.  Unable to verify token.", status=404)
```
ajmaddaford: Why don't all instances of 4XX status codes use ClientError?

ras_party/controllers/account_controller.py
ajmaddaford: I think we need to be careful about trying to be too helpful in the log messages and repeating information already in the log message. For example here, the context of the respondent not existing should be clear if we're logging the url, without needing to add "Unable to change respondent's password"

ras_party/controllers/account_controller.py
ajmaddaford: I think this log message is a little misleading - we don't send emails from the party service. We should be explicit in our log messages, something like "Error sending request to notify gateway" would be clearer here.

ajmaddaford (changes requested):

ras_party/controllers/account_controller.py
gavinleeedwards: I don't necessarily agree that being too helpful is a bad thing.  The issue here is that `Respondent does not exist` could come from several places in the code so making sort of messages less generic makes it quicker to pinpoint where any issues might be occurring.

ras_party/controllers/account_controller.py
gavinleeedwards: It depends on the circumstances.  For example, all `Respondent does not exist` errors are 404s.  In some cases, it doesn't matter if the exception is raised so it's a ClientError (e.g. if a user mistypes a respondemt id).  In other cases, it's an issue if the respondent isn't found if we're expecting it to, so RasError.

ajmaddaford: I think by doing this we're choosing which exception to raise based on what log level we want the message output at. This would be better dealt with by the service calling this endpoint to interpret the result and how to log it.

Also, by raising it as a `RasError` it's logged as an exception with the text 'Uncaught exception'.

ras_party/controllers/account_controller.py
ajmaddaford: I think we'll end up with a lot of duplication. 'Expired email verification token' and 'Unknown email verification token' appear multiple times but haven't had any more logging added. If we applied the same 'Unable to change respondent's password' to all of those log lines there would be a lot of repetition. If the url was being logged we'd see 'Respondent does not exist' along with `url=/respondents/change_password/<token>, making it clear what action was being performed when the error occurred.

gavinleeedwards:
@ajmaddaford I've changed all 400s to ClientError, added the url to log messages, changed the RasNotify error message and made sure that the kwargs are now passed through to the exception logs. 👍

ras_party/support/session_decorator.py
```diff
@@ -21,12 +21,16 @@ def handle_session(f, args, kwargs):
         logger.info("Committing database session.")
         session.commit()
         return result
+    except ClientError:
+        logger.info(f"Rolling back database session due to a ClientError")
+        session.rollback()
+        raise
     except RasError:
-        logger.error(f"Rolling back database session due to failure executing function")
+        logger.info(f"Rolling back database session due to failure executing function")
```
ajmaddaford: Minor, but this doesn't need to be an f-string

ras_party/exceptions.py
```diff
@@ -1,3 +1,15 @@
+class ClientError(Exception):
+
+    status_code = 400
+
+    def __init__(self, errors, status=None, **kwargs):
+        self.errors = errors if isinstance(errors, list) else [errors]
+        self.status_code = status or RasPartyError.status_code
```
ajmaddaford: This will set the default error code to 500 (`RasPartyError` is a `RasError`). If we defaulted status to 400 rather than `None` we could drop the `status_code = 400`:
```python
class ClientError(Exception):

    def __init__(self, errors, status=400, **kwargs):
        self.errors = errors if isinstance(errors, list) else [errors]
        self.status_code = status
        self.kwargs = kwargs
```

ras_party/controllers/account_controller.py
```diff
@@ -155,7 +158,8 @@ def change_respondent_enrolment_status(payload, session):
                 status=change_flag)
     respondent = query_respondent_by_party_uuid(respondent_id, session)
     if not respondent:
-        raise RasError("Respondent does not exist.", status=404)
+        raise ClientError("Respondent does not exist.  Unable to change enrolment status",
```
ajmaddaford: I think we can drop the "Unable to change enrolment status" now that we're logging the url

ras_party/controllers/account_controller.py
```diff
@@ -311,7 +316,8 @@ def change_respondent_account_status(payload, party_id, session):
 
     respondent = query_respondent_by_party_uuid(party_id, session)
     if not respondent:
-        raise RasError("Unable to find respondent account", status=404)
+        raise ClientError("Respondent does not exist.  Unable to change account status.", respondent_id=party_id,
```
ajmaddaford: I think we can drop the "Unable to change account status" now that we're logging the url

ras_party/controllers/account_controller.py
```diff
@@ -58,18 +58,20 @@ def post_respondent(party, tran, session):
...
 
     iac = request_iac(party['enrolmentCode'])
     if not iac.get('active'):
-        raise RasError("Enrolment code is not active.", status=400)
+        raise ClientError("Enrolment code is not active.", status=400)
 
     existing = query_respondent_by_email(party['emailAddress'], session)
     if existing:
-        raise RasError("Email address already exists.", status=400, party_uuid=party.get('party_uuid', None))
+        raise ClientError("Email address already exists.",
```
ajmaddaford: The existing log lines are not consistent on whether they end with a full stop or not. I say they shouldn't as the message should be a single event. For any we're amending, I think we should drop the trailing full stop.

ras_party/controllers/account_controller.py
```diff
@@ -257,7 +262,7 @@ def change_respondent_password(token, payload, tran, session):
         NotifyGateway(current_app.config).confirm_password_change(
             email_address, personalisation, str(party_id))
     except RasNotifyError:
-        logger.error('Error sending notification email', party_id=party_id)
+        logger.error('Error sending notification email', respondent_id=party_id)
 
     # This ensures the log message is only written once the DB transaction is committed
     tran.on_success(lambda: logger.info('Respondent has changed their password', party_id=party_id))
```
ajmaddaford: This log line is still `party_id`, should it be updated to `respondent_id` as above?

ras_party/controllers/account_controller.py
```diff
@@ -297,7 +302,7 @@ def request_password_change(payload, session):
                 email_address, personalisation, str(party_id))
         except RasNotifyError:
             # Note: intentionally suppresses exception
-            logger.error('Error sending notification email for party_id', party_id=party_id)
+            logger.error('Error sending request to Notify Gateway for respondent_id', respondent_id=party_id)
```
ajmaddaford: The " for respondent_id" isn't needed in the log event, as it will be clear from the log line including `respondent_id`.

ajmaddaford (changes requested):

ras_party/support/session_decorator.py
gavinleeedwards: Done 👍

ras_party/exceptions.py
gavinleeedwards: Done 👍

ras_party/controllers/account_controller.py
gavinleeedwards: Done 👍

ras_party/controllers/account_controller.py
gavinleeedwards: Done 👍

ras_party/controllers/account_controller.py
gavinleeedwards: Done - removed all full stops to keep log messages consistent.

ras_party/controllers/account_controller.py
gavinleeedwards: Done 👍

ras_party/controllers/account_controller.py
gavinleeedwards: Done 👍

gavinleeedwards (commented):
@ajmaddaford All done and ready for re-review 👍

ras_party/controllers/account_controller.py
```diff
@@ -269,13 +274,13 @@ def change_respondent_password(token, payload, tran, session):
...
+        raise ClientError(v.errors, 400)
 
     email_address = payload['email_address']
 
     respondent = query_respondent_by_email(email_address, session)
     if not respondent:
-        raise RasError("Respondent does not exist.", status=404)
+        raise ClientError("Respondent does not exist", status=404)
 
     logger.debug("Requesting password change", party_id=respondent.party_uuid)
```
ajmaddaford, line 285: This appears in the log as:
```
{"party_id": "UUID('b73ad0e0-c83d-454b-93ac-12ed58c741e2')", "event": "Requesting password change", ... }
```
We should wrap `respondent.party_uuid` with `str` in all places where we're accessing the uuid from the database object.

ras_party/error_handlers.py
```diff
@@ -2,21 +2,37 @@
...
+
 @blueprint.app_errorhandler(RasError)
 def ras_error(error):
     response = jsonify(error.to_dict())
     response.status_code = error.status_code
-    logger.exception('Uncaught exception', errors=error.to_dict(), status=error.status_code)
+    logger.exception('Uncaught exception',
+                     url=request.url,
+                     errors=error.to_dict(),
+                     **error.kwargs,
```
ajmaddaford: Minor, but I think `**error.kwargs` should be the last argument, consistent with how it's used in function definitions.

ras_party/controllers/account_controller.py
```diff
@@ -294,17 +294,18 @@ def request_password_change(payload, session):
...
 
-        logger.info('Reset password url', url=verification_url, party_id=respondent.party_uuid)
+        party_id = str(respondent.party_uuid)
+
+        logger.info('Reset password url', url=verification_url, party_id=party_id)
 
-        party_id = respondent.party_uuid
         try:
             NotifyGateway(current_app.config).request_password_change(
                 email_address, personalisation, str(party_id))
```
ajmaddaford: `party_id` is now a string so we can drop the `str` here

ras_party/controllers/account_controller.py
```diff
@@ -89,7 +89,7 @@ def post_respondent(party, tran, session):
     business = query_business_by_party_uuid(business_id, session)
     if not business:
         raise ClientError("Could not locate business when creating business association",
-                          business_id=business_id,
+                          business_id=str(business_id),
```
ajmaddaford: `business_id` came from payload, so should already be a string

ras_party/controllers/account_controller.py
```diff
@@ -159,7 +159,7 @@ def change_respondent_enrolment_status(payload, session):
     respondent = query_respondent_by_party_uuid(respondent_id, session)
     if not respondent:
         raise ClientError("Respondent does not exist",
-                          respondent_id=respondent_id, status=404)
+                          respondent_id=str(respondent_id), status=404)
```
ajmaddaford: `respondent_id ` came from payload, so should already be a string

ras_party/controllers/account_controller.py
```diff
@@ -316,7 +317,7 @@ def change_respondent_account_status(payload, party_id, session):
 
     respondent = query_respondent_by_party_uuid(party_id, session)
     if not respondent:
-        raise ClientError("Respondent does not exist", respondent_id=party_id,
+        raise ClientError("Respondent does not exist", respondent_id=str(party_id),
```
ajmaddaford: `respondent_id ` comes from the url, so should already be a string

ras_party/controllers/account_controller.py
```diff
@@ -410,7 +411,7 @@ def resend_verification_email(party_uuid, session):
     :param party_uuid: the party uuid
     :return: make_response
     """
-    logger.debug('Attempting to resend verification_email', party_uuid=party_uuid)
+    logger.debug('Attempting to resend verification_email', party_uuid=str(party_uuid))
```
ajmaddaford: `party_uuid ` comes from the url, so should already be a string

ras_party/controllers/account_controller.py
```diff
@@ -462,7 +463,7 @@ def add_new_survey_for_respondent(payload, tran, session):
         business = query_business_by_party_uuid(business_id, session)
         if not business:
             raise ClientError("Could not locate business when creating business association",
-                              business_id=business_id,
+                              business_id=str(business_id),
```
ajmaddaford: `business_id` comes from the response from the case service, so should already be a string

ras_party/controllers/account_controller.py
```diff
@@ -477,26 +478,26 @@ def add_new_survey_for_respondent(payload, tran, session):
     # This ensures the log message is only written once the DB transaction is committed
     tran.on_success(lambda: logger.info('Respondent has enroled to survey for business',
                                         survey_name=survey_name,
-                                        business=business_id))
+                                        business=str(business_id)))
```
ajmaddaford: `business_id` comes from the response from the case service, so should already be a string

ras_party/controllers/business_controller.py
```diff
@@ -67,7 +67,7 @@ def get_business_by_id(party_uuid, session, verbose=False, collection_exercise_i
 
     business = query_business_by_party_uuid(party_uuid, session)
     if not business:
-        raise ClientError("Business with party id does not exist", party_uuid=party_uuid, status=404)
+        raise ClientError("Business with party id does not exist", party_uuid=str(party_uuid), status=404)
```
ajmaddaford: `party_uuid` comes from the url, so should already be a string

ras_party/controllers/party_controller.py
```diff
@@ -58,12 +58,12 @@ def get_party_by_id(sample_unit_type, id, session):
     if sample_unit_type == Business.UNIT_TYPE:
         business = query_business_by_party_uuid(id, session)
         if not business:
-            raise ClientError("Business with id does not exist", business_id=id, status=404)
+            raise ClientError("Business with id does not exist", business_id=str(id), status=404)
```
ajmaddaford: `id` comes from the url, so should already be a string

ras_party/controllers/party_controller.py
```diff
@@ -58,12 +58,12 @@ def get_party_by_id(sample_unit_type, id, session):
...
         return business.to_party_dict()
     elif sample_unit_type == Respondent.UNIT_TYPE:
         respondent = query_respondent_by_party_uuid(id, session)
         if not respondent:
-            raise ClientError("Respondent with id does not exist", respondent_id=id, status=404)
+            raise ClientError("Respondent with id does not exist", respondent_id=str(id), status=404)
```
ajmaddaford: `id` comes from the url, so should already be a string

ras_party/controllers/respondent_controller.py
```diff
@@ -78,7 +78,7 @@ def change_respondent_details(respondent_data, respondent_id, session):
 
     respondent = query_respondent_by_party_uuid(respondent_id, session)
     if not respondent:
-        raise ClientError("Respondent id does not exist", respondent_id=respondent_id, status=404)
+        raise ClientError("Respondent id does not exist", respondent_id=str(respondent_id), status=404)
```
ajmaddaford: `respondent_id ` comes from the url, so should already be a string

ajmaddaford (changes requested):

ras_party/controllers/respondent_controller.py
gavinleeedwards: @ajmaddaford  All done.

ajmaddaford (approved):

PR #152, opened 2018-08-20
Remove references of survey_name across service

MIslam92:
### What is the Context of this PR?

In context of this PR is very similar to: https://github.com/ONSdigital/rm-collection-exercise-service/pull/113

Since the service name can be edited now, the one held in this service is not even the correct one i.e. out of date and the actual up to date survey details are stored in another service i.e. survey service

Card: https://trello.com/c/LP687BIU/331-remove-repeated-survey-name-from-services-other-than-survey-service-s

This PR aims to remove any ref of the survey name and drop the column entirely from the DB as well by adding a new script.

### How to Review
- Verify all instances of survey name are removed and any code referencing it.
- Check all tests pass and that the new alembic script works. 
- Ensure all functionality works as per norm

### DIFF
--- a/migrations/versions/613bf3a40948_remove_survey_name.py
+++ b/migrations/versions/613bf3a40948_remove_survey_name.py
@@ -1,0 +1,23 @@
+"""remove survey name
+
+Revision ID: 613bf3a40948
+Revises: ac604d2cf763
+Create Date: 2018-08-20 10:16:35.995473
+
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '613bf3a40948'
+down_revision = 'ac604d2cf763'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    pass
+
+
+def downgrade():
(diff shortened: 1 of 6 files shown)


shahi645 (approved):
LGTM

ryangrundy7 (approved):
LGTM 👍

migrations/versions/613bf3a40948_remove_survey_name.py
```diff
@@ -0,0 +1,23 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    pass
+
+
+def downgrade():
+    op.drop_column('survey_name', schema='partysvc.enrolment')
```
andrewmil, line 23: Is this right? I would have thought removing the column would be in the upgrade section?

MIslam92: Good spot.
Just created a PR fix for it: https://github.com/ONSdigital/ras-party/pull/168
It should fix it.

PR #153, opened 2018-08-20
Remove bi/modify case events and call iac directly

gemmai95:
# Motivation and Context

This change is required for the removal of BI cases. Stops case service getting called unnecessarily. 

# What has changed

Added ability for iac code to get disabled by service when respondent is creating an account or adding a survey. Only call case service if first respondent to enrol for survey and business.

Added IAC controller to handle requests to IAC service.

# How to test?

Tested with changes in case service and acceptance tests.

# Links

Trello:
https://trello.com/c/V8twXPlu

PRs:
https://github.com/ONSdigital/rm-collection-exercise-service/pull/108
https://github.com/ONSdigital/rm-case-service/pull/70
https://github.com/ONSdigital/rasrm-acceptance-tests/pull/165
https://github.com/ONSdigital/rm-action-service/pull/79

### DIFF
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -324,6 +324,26 @@ def request_password_change(payload, session):
 
 
 @with_db_session
+def resend_password_email_expired_token(token, session):
+    """
+    Check and resend an email verification email using the expired token
+    :param token: the expired token
+    :param session: database session
+    :return: response
+    """
+    email_address = decode_email_token(token, duration=None)
+    respondent = query_respondent_by_email(email_address, session)
+
+    if not respondent:
+        raise ClientError("Respondent does not exist", status=404)
+
+    payload = {'email_address': email_address}
+
+    response = request_password_change(payload)
+    return response
+
+
+@with_db_session
 def notify_change_account_status(payload, party_id, session):
     status = payload['status_change']
 
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -171,7 +171,6 @@ def change_respondent_enrolment_status(payload, session):
         logger.info('No active enrolments', business_id=business_id, survey_id=survey_id)
         for case in get_cases_for_casegroups(casegroup_ids, business_id):
             post_case_event(case['id'],
-                            party_id=None,
                             category='NO_ACTIVE_ENROLMENTS',
                             desc='No active enrolments remaining for case')
 
@@ -556,7 +555,7 @@ def add_new_survey_for_respondent(payload, tran, session):
     if count_enrolment_by_survey_business(survey_id, business_id, session) == 0:
         casegroup_ids = get_business_survey_casegroups(survey_id, business_id)
         for case in get_cases_for_casegroups(casegroup_ids, business_id):
-            post_case_event(str(case['id']), None, "RESPONDENT_ENROLED", "Respondent enroled")
+            post_case_event(case['id'], "RESPONDENT_ENROLED", "Respondent enroled")
 
     # This ensures the log message is only written once the DB transaction is committed
     tran.on_success(lambda: logger.info('Respondent has enroled to survey for business',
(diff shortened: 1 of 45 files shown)


andrewmil (changes requested):
Changes look good, tested with remove-bi collection exericse, case and party PR's. Most case events working except `DISABLE_RESPONDENT_ENROLMENT` mentioned in the case PR.

Getting linting errors for `make test`
```
./ras_party/controllers/account_controller.py:540:121: E501 line too long (127 > 120 characters)
```

ras_party/controllers/account_controller.py
```diff
@@ -536,7 +534,9 @@ def enrol_respondent_for_survey(r, sess):
     # Send an enrolment event to the case service
     case_id = pending_enrolment.case_id
     logger.info('Pending enrolment for case_id', case_id=case_id)
-    post_case_event(str(case_id), str(r.party_uuid), "RESPONDENT_ENROLED", "Respondent enrolled")
+    if count_enrolment_by_survey_business(survey_id=enrolment.survey_id, business_id=enrolment.business_id,
+                                          session=sess) == 0:
+        post_case_event(str(case_id), None, "RESPONDENT_ENROLED", "Respondent enrolled")
```
andrewmil: Untested

andrewmil (approved):
Looks good, changes appear to work with other remove-bi PR's :+1:

ras_party/controllers/account_controller.py
```diff
@@ -69,6 +70,8 @@ def post_respondent(party, tran, session):
     if not iac.get('active'):
         raise ClientError("Enrolment code is not active", status=400)
 
+    disable_iac(party['enrolmentCode'])
```
ajmaddaford: - The documentation for this `post_respondent` method needs to be updated now that we're no longer creating a case event and are disabling the enrolment code.
- If there is a failure when committing the database records does that mean the respondent would be unable to retry (as the enrolment code would be disabled)?

ajmaddaford (changes requested):

ras_party/controllers/account_controller.py
gemmai95: @ajmaddaford I will update the documentation. As for the issue with disabling the iac this is the same behaviour that would currently happen. I can have a look at moving this to after we commit to the database.

ajmaddaford: The behaviour is definitely different now. I followed these steps:
- Attempt to create an account with an existing email address
- On the validation error for the duplicate email address, change it to one that doesn't already exist and submit

On concourse-latest it works, on this branch a 500 error is generated.

ras_party/controllers/account_controller.py
```diff
@@ -135,7 +130,7 @@ def post_respondent(party, tran, session):
         raise RasError("Error during enrolment process", status=500)
 
     register_user(party, tran)
-
+    disable_iac(enrolment_code=party['enrolmentCode'])
```
ajmaddaford: - Now that the session is committed above, we should stop passing `tran` to `register_user` and remove the `dummy_compensating_action` code from that method.
- What should the behaviour be if we have successfully enrolled and registered a user, but disabling the IAC fails? I think currently this will create a 500 error for the user.

ras_party/controllers/account_controller.py
```diff
@@ -495,8 +488,12 @@ def add_new_survey_for_respondent(payload, tran, session):
                           survey_id=survey_id,
                           status=EnrolmentStatus.ENABLED)
     session.add(enrolment)
+    session.commit()
 
-    post_case_event(str(case_id), str(respondent_party_id), "RESPONDENT_ENROLED", "Respondent enroled")
+    disable_iac(enrolment_code)
```
ajmaddaford: Same as comment above for behaviour if disabling the enrolment code fails.

ajmaddaford (changes requested):

ras_party/controllers/account_controller.py
gemmai95: @ajmaddaford I spoke to Dave and he basically just said log that the iac has failed to be deactivated using the case id and a splunk alert should probably be set up to catch this. Do we want the process to carry on after as the respondent will have been added to django and party?

ajmaddaford: 👍 that sounds reasonable

ras_party/controllers/account_controller.py
```diff
@@ -458,7 +448,7 @@ def put_email_verification(token, tran, session):
     return respondent.to_respondent_dict()
 
 
-def update_verified_email_address(respondent, tran, session):
+def update_verified_email_address(respondent, tran, _):
```
ajmaddaford: The session parameter should be removed if it's not being used.

ajmaddaford (approved):

PR #157, opened 2018-08-22
Move launching of docker containers to tox

jcox-dev:
`make test` now brings up and tears down the required db.


jcox-dev:
No idea why this is failing on travis but closing for now as it's not hugely necessary.

PR #158, opened 2018-08-22
Replace Explicit Requirements With *

PurpleBooth:
Motivation and Context
==

This seems to be the only service with explicit pinning of versions. This will
change the explicit pinning to be "*" pinning like the other services

What has changed
==

Locked versions are still exactly the same, only the Pipfile has changed.

How to test?
==

Automated tests.

Links
==


mfcrocker (approved):

jcox-dev (approved):
Packages all install fine and tests pass ✨

PR #167, opened 2018-08-28
Added account controller to unlock account from party service

shahi645:
# Motivation and Context
We want party to call out to django service, when an internal user unlocks an account in ROps. 

# What has changed
Simple functionality added in to existing method to just call out. 

# How to test?
Needs to be run with rest of the changes found on this ticket:
https://trello.com/c/mm4SZ3N5/268-us203-int-ability-to-unlock-a-respondents-account-l

### DIFF
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -333,6 +333,14 @@ def change_respondent_account_status(payload, party_id, session):
     if not respondent:
         raise ClientError("Respondent does not exist", respondent_id=party_id,
                           status=404)
+    # Unlock respondents account
+    if status == 'ACTIVE':
+        email_address = respondent.email_address
+        oauth_response = OauthClient().update_account(username=email_address, account_locked='False')
+
+        if oauth_response.status_code != 201:
+            raise RasError('Failed to unlock respondent account', respondent_id=str(respondent.party_uuid))
+
     respondent.status = status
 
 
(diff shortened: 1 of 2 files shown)


ras_party/controllers/account_controller.py
```diff
@@ -331,6 +331,14 @@ def change_respondent_account_status(payload, party_id, session):
     if not respondent:
         raise ClientError("Respondent does not exist", respondent_id=party_id,
                           status=404)
+    # Unlock respondents account
+    if status == 'ACTIVE':
+        email_address = respondent.email_address
+        oauth_response = OauthClient().update_account(username=email_address, account_locked='False')
+
+        if oauth_response.status_code != 201:
```
atrozzo: what about use raise_for_status ?

atrozzo: Do youhave any unit test for this code ?

atrozzo (changes requested):

ras_party/controllers/account_controller.py
shahi645: i did think `raise_for_status` would be best practice, but i followed the pattern of the file. And it is tested through another integration style test

atrozzo: Sorry but there is not such thing as pattern of a file, you should chose teh best option or practice regardless what other developers more or lesse experinced do . Not sure that test_put_change_respondent_account_status_active is actually testing something. Plus you added a new condition checking status code != 201 so u should test you raise RasError.

shahi645: ok sure ill use `raise_for_status` and will double check the tests

tej99 (approved):
LGTM

MIslam92 (approved):

PR #168, opened 2018-08-29
Fix to ensure survey_name is removed from db

MIslam92:
### Context
Fix for: https://github.com/ONSdigital/rm-collection-exercise-service/pull/113

### How to Test
Actually check its been removed from the db i.e. survey_name column


andrewmil (approved):

shahi645 (approved):
Tested Db drops column 👍

PR #170, opened 2018-09-06
Fix to ensure right column is removed and correct schema referenced

MIslam92:
### Context
Fix for: https://github.com/ONSdigital/ras-party/pull/168

### How to Test
Actually check its been removed from the db i.e. survey_name column


andrewmil (approved):
LGTM :+1:

PR #172, opened 2018-09-14
206 resend verification if using expired link

gavinleeedwards:
# Motivation and Context
If an expired verification link is attempted, we need to give the external user an opportunity to resend the verification link to their registered email address.  When they attempt the link, all we know is the token so there needs to be a new way to be able to resend a link using only the token.

# What has changed
- New endpoint which will be called from frontstage and given the expired token.
- Added tests.

# How to test?
- Set `EMAIL_TOKEN_EXPIRY` to 0 so that tokens are immediately expired (otherwise you'll be waiting 80 hours for it to expire).
- Run alongside branch `206-resend-verification-if-using-expired-link` in ras-frontstage.
- A new verification link should be generated when the new endpoint is hit from frontstage (check the logs when testing).
- All tests should pass.

# Links
- https://trello.com/c/fqc35BzO

### DIFF
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -669,3 +669,15 @@ def get_cases_for_casegroups(casegroup_ids, case_party_id):
     logger.debug('Successfully retrieved cases for casegroups',
                  case_party_id=case_party_id)
     return matching_cases
+
+
+@with_db_session
+def resend_verification_email_expired_token(token, session):
+    email_address = decode_email_token(token, duration=None)
+    respondent = query_respondent_by_email(email_address, session)
+
+    if not respondent:
+        raise ClientError("Respondent does not exist", status=404)
+
+    response = resend_verification_email(respondent.party_uuid)
+    return response
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -417,10 +417,11 @@ def update_verified_email_address(respondent, tran, session):
 
 
 @with_db_session
-def resend_verification_email(party_uuid, session):
+def resend_verification_email_by_uuid(party_uuid, session):
     """
-    Check and resend an email verification email
+    Check and resend an email verification email using the party id
     :param party_uuid: the party uuid
+    :param session: database session
     :return: make_response
     """
     logger.debug('Attempting to resend verification_email', party_uuid=party_uuid)
@@ -429,10 +430,33 @@ def resend_verification_email(party_uuid, session):
     if not respondent:
         raise ClientError(NO_RESPONDENT_FOR_PARTY_ID, status=404)
 
+    response = _resend_verification_email(respondent)
+    return response
+
+
+@with_db_session
+def resend_verification_email_expired_token(token, session):
+    """
+    Check and resend an email verification email using the expired token
+    :param token: the expired token
+    :param session: database session
+    :return: make_response
+    """
+    email_address = decode_email_token(token, duration=None)
+    respondent = query_respondent_by_email(email_address, session)
+
+    if not respondent:
+        raise ClientError("Respondent does not exist", status=404)
+
+    response = _resend_verification_email(respondent)
+    return response
+
+
+def _resend_verification_email(respondent):
     if respondent.pending_email_address:
-        _send_email_verification(party_uuid, respondent.pending_email_address)
+        _send_email_verification(respondent.party_uuid, respondent.pending_email_address)
     else:
-        _send_email_verification(party_uuid, respondent.email_address)
+        _send_email_verification(respondent.party_uuid, respondent.email_address)
 
     return {'message': EMAIL_VERIFICATION_SENT}
 
@@ -669,15 +693,3 @@ def get_cases_for_casegroups(casegroup_ids, case_party_id):
     logger.debug('Successfully retrieved cases for casegroups',
                  case_party_id=case_party_id)
     return matching_cases
-
-
-@with_db_session
-def resend_verification_email_expired_token(token, session):
-    email_address = decode_email_token(token, duration=None)
-    respondent = query_respondent_by_email(email_address, session)
-
-    if not respondent:
-        raise ClientError("Respondent does not exist", status=404)
-
-    response = resend_verification_email(respondent.party_uuid)
-    return response
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -422,7 +422,7 @@ def resend_verification_email_by_uuid(party_uuid, session):
     Check and resend an email verification email using the party id
     :param party_uuid: the party uuid
     :param session: database session
-    :return: make_response
+    :return: response
     """
     logger.debug('Attempting to resend verification_email', party_uuid=party_uuid)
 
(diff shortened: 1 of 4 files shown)


ras_party/controllers/account_controller.py
```diff
@@ -669,3 +669,15 @@ def get_cases_for_casegroups(casegroup_ids, case_party_id):
...
+
+@with_db_session
+def resend_verification_email_expired_token(token, session):
+    email_address = decode_email_token(token, duration=None)
+    respondent = query_respondent_by_email(email_address, session)
+
+    if not respondent:
+        raise ClientError("Respondent does not exist", status=404)
+
+    response = resend_verification_email(respondent.party_uuid)
```
AndrewTorrance: Not sure why we are going via resend_verification_email , this does a query for respondent by UUID which I THINK is redundant if we have already got got the respondent by email ?  If my assumption is correct then could we split out the last if else into its own private function and call it from both resend_verification_email_expired_token and resend_verification_email ( also possibly change the name of resend_verification_email to resend_verification_email_by_uuid ?)

AndrewTorrance (changes requested):
I think we need a test of email not found 404 ? Also left comments where I think we are doing one more db look up than we need to because we are using resend_verification_email to send

ras_party/controllers/account_controller.py
```diff
@@ -417,10 +417,11 @@ def compensate_oauth_change():
...
 
 @with_db_session
-def resend_verification_email(party_uuid, session):
+def resend_verification_email_by_uuid(party_uuid, session):
     """
-    Check and resend an email verification email
+    Check and resend an email verification email using the party id
     :param party_uuid: the party uuid
+    :param session: database session
     :return: make_response
```
shahi645: should just be response

ras_party/controllers/account_controller.py
```diff
@@ -429,10 +430,33 @@ def resend_verification_email(party_uuid, session):
...
+    return response
+
+
+@with_db_session
+def resend_verification_email_expired_token(token, session):
+    """
+    Check and resend an email verification email using the expired token
+    :param token: the expired token
+    :param session: database session
+    :return: make_response
```
shahi645: same as above

test/test_respondent_controller.py
```diff
@@ -311,6 +311,25 @@ def test_resend_verification_email_sends_via_notify(self):
...
+        # When the resend verification with expired token endpoint is hit
+        self.resend_verification_email_expired_token(token)
+        # Then a notification is sent the the respondent's email adddress
+        self.assertTrue(self.mock_notify.verify_email.called)
+
+    def test_resend_verification_email_expired_token_respondent_not_found(self):
+        # The token is valid but the respondent doesn't exist
+        current_app.config['EMAIL_TOKEN_EXPIRY'] = -1
+        token = self.generate_valid_token_from_email('invalid@email.com')
+        #When the resend verification with expired token endpoint is hit
```
shahi645: add a space here between `#` and `When` otherwise linting fail

shahi645 (changes requested):

ras_party/controllers/account_controller.py
gavinleeedwards: Changed :+1:

ras_party/controllers/account_controller.py
gavinleeedwards: Changed :+1:

test/test_respondent_controller.py
gavinleeedwards: It did fail the linting.  Updated that now.

ras_party/controllers/account_controller.py
gavinleeedwards: Updated to remove the second unnecessary call to the db.

ras_party/views/account_view.py
```diff
@@ -73,13 +73,13 @@ def put_email_verification(token):
...
-@account_view.route('/resend-verification-email/<party_uuid>', methods=['GET'])
+@account_view.route('/resend-verification-email/<party_uuid>', methods=['POST'])
 def resend_verification_email(party_uuid):
     response = account_controller.resend_verification_email_by_uuid(party_uuid)
     return make_response(jsonify(response), 200)
 
 
-@account_view.route('/resend-verification-email-expired-token/<token>', methods=['GET'])
+@account_view.route('/resend-verification-email-expired-token/<token>', methods=['POST'])
 def resend_verification_email_expired_token(token):
```
shahi645: I'm not sure if this should be a `POST`? We are not changing any data? But maybe worth double checking with Lewis

shahi645 (changes requested):

shahi645:
@gavinleeedwards I have the same codacy issue in my PR, do we just ignore?

ras_party/controllers/account_controller.py
```diff
@@ -417,22 +417,46 @@ def compensate_oauth_change():
...
     :param party_uuid: the party uuid
-    :return: make_response
+    :param session: database session
+    :return: response
     """
     logger.debug('Attempting to resend verification_email', party_uuid=party_uuid)
 
     respondent = query_respondent_by_party_uuid(party_uuid, session)
     if not respondent:
         raise ClientError(NO_RESPONDENT_FOR_PARTY_ID, status=404)
```
shahi645, line 445: This line isnt tested, but i guess its probably tested for the other methods

ras_party/controllers/account_controller.py
```diff
@@ -417,22 +417,46 @@ def compensate_oauth_change():
 
 
 @with_db_session
-def resend_verification_email(party_uuid, session):
+def resend_verification_email_by_uuid(party_uuid, session):
```
shahi645, line 434: 👍

ras_party/controllers/account_controller.py
```diff
@@ -417,22 +417,46 @@ def compensate_oauth_change():
...
+    respondent = query_respondent_by_email(email_address, session)
+
+    if not respondent:
+        raise ClientError("Respondent does not exist", status=404)
+
+    response = _resend_verification_email(respondent)
+    return response
+
+
+def _resend_verification_email(respondent):
```
shahi645, line 469: Nice use of common functionality, we should do this more often in our services 👍

ras_party/views/account_view.py
```diff
@@ -73,9 +73,15 @@ def put_email_verification(token):
     return make_response(jsonify(response), 200)
 
 
-@account_view.route('/resend-verification-email/<party_uuid>', methods=['GET'])
+@account_view.route('/resend-verification-email/<party_uuid>', methods=['POST'])
```
ryangrundy7, line 76: Did you mean to change this from a `GET` to a `POST`?

gavinleeedwards: These should be posts but I needed to change frontstage to match.  All done now.

ras_party/views/account_view.py
gavinleeedwards: Kept as a POST request and changed frontstage to be in line.

shahi645 (approved):

ryangrundy7 (approved):

AndrewTorrance (approved):

PR #175, opened 2018-09-24
207 resend password reset link if expired

gavinleeedwards:
# Motivation and Context
When a respondent requests a change of password, they are emailed a link that expires after 80 hours.  If they try to use the link after the 80 hours then they are redirected to a page that tells them to phone us for another link.  This PR allows the respondent to generate a new link which is then sent to the email registered to their account.

# What has changed

New endpoint which takes an expired token, decodes it, generates a new token and sends it to the registered email address of the user.

# How to test?
- Set the EMAIL_TOKEN_EXPIRY to -1 so that tokens expire as soon as they're created (otherwise you'll be waiting 80 hours for one to expire)
- Run this branch alongside `207-resend-password-reset-link-if-expired` in frontstage
- In frontstage, request a password reset for an active account - check the party logs to find it
- Follow the link to reset the password.  The link should be expired and you should be redirected to a page and given the option to resend a password reset link
- Click the link to send another and check in the party logs that another link has been generated.
- You should finally be redirected to the 'Check your email page'
- The new link that's been generated in party should allow you to reset your password as normal (if you change the EMAIL_TOKEN_EXPIRY back)

# Links
- https://trello.com/c/1FXbQiA2


shahi645 (approved):
Changes look simple and good to me 🌮

ryangrundy7 (approved):

PR #176, opened 2018-09-25
Replace generic 'RasError' and 'ClientError'

tej99:
# Motivation and Context

Nearly every exception raised in this service is a 'RasError' or 'ClientError' exception (which is an extension of RasError).
Every RasError raised was logged at error level and with the message 'Uncaught error'.

# What has changed

I have removed these generic exceptions and replaced them with the appropriate HTTPException.
Logging has also been improved to only have logging at error level when necessary.

# How to test?

Ran integration tests
and AT

# Links

https://trello.com/c/9Fc16htf/279-ras-party-logging-and-exceptions-m

### DIFF
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -5,7 +5,7 @@ import structlog
 from flask import current_app
 from itsdangerous import SignatureExpired, BadSignature, BadData
 from requests import HTTPError
-from sqlalchemy import orm
+from werkzeug.exceptions import BadRequest, Conflict, InternalServerError,  NotFound
 
 from ras_party.clients.oauth_client import OauthClient
 from ras_party.controllers.notify_gateway import NotifyGateway
@@ -15,7 +15,7 @@ from ras_party.controllers.queries import query_enrolment_by_survey_business_res
 from ras_party.controllers.queries import query_respondent_by_email, query_respondent_by_pending_email
 from ras_party.controllers.queries import query_respondent_by_party_uuid, query_business_by_party_uuid
 from ras_party.controllers.validate import Exists, Validator
-from ras_party.exceptions import ClientError, RasError, RasNotifyError
+from ras_party.exceptions import RasNotifyError
 from ras_party.models.models import BusinessRespondent, Enrolment, EnrolmentStatus
 from ras_party.models.models import PendingEnrolment, Respondent, RespondentStatus
 from ras_party.support.public_website import PublicWebsite
@@ -61,37 +61,35 @@ def post_respondent(party, tran, session):
         try:
             uuid.UUID(party['id'])
         except ValueError:
-            raise ClientError(f"'{party['id']}' is not a valid UUID format for property 'id'", status=400)
+            logger.debug("Invalid respondent id type", respondent_id=party['id'])
+            raise BadRequest(f"'{party['id']}' is not a valid UUID format for property 'id'")
 
     if not v.validate(party):
-        raise ClientError(v.errors, 400)
+        logger.debug(v.errors)
+        raise BadRequest(v.errors)
 
     iac = request_iac(party['enrolmentCode'])
     if not iac.get('active'):
-        raise ClientError("Enrolment code is not active", status=400)
+        logger.debug("Inactive enrolment code")
+        raise BadRequest("Enrolment code is not active")
 
     existing = query_respondent_by_email(party['emailAddress'], session)
     if existing:
-        raise ClientError("Email address already exists",
-                          status=400,
-                          party_uuid=str(existing.party_uuid))
+        logger.debug("Email already exists", party_uuid=str(existing.party_uuid))
+        raise BadRequest("Email address already exists")
 
     case_context = request_case(party['enrolmentCode'])
     case_id = case_context['id']
     business_id = case_context['partyId']
     collection_exercise_id = case_context['caseGroup']['collectionExerciseId']
     collection_exercise = request_collection_exercise(collection_exercise_id)
-
-    try:
-        survey_id = collection_exercise['surveyId']
-    except KeyError:
-        raise ClientError("There is no survey bound for this user", party_uuid=party['party_uuid'])
+    survey_id = collection_exercise['surveyId']
 
     business = query_business_by_party_uuid(business_id, session)
     if not business:
-        raise ClientError("Could not locate business when creating business association",
-                          business_id=business_id,
-                          status=404)
+        logger.error("Could not locate business when creating business association",
+                     business_id=business_id, case_id=case_id, collection_exercise_id=collection_exercise_id)
+        raise InternalServerError("Could not locate business when creating business association")
 
     # Chain of enrolment processes
     translated_party = {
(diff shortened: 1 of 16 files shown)


ras_party/controllers/account_controller.py
```diff
@@ -85,13 +88,13 @@ def post_respondent(party, tran, session):
     try:
         survey_id = collection_exercise['surveyId']
     except KeyError:
-        raise ClientError("There is no survey bound for this user", party_uuid=party['party_uuid'])
+        logger.debug("No survey bound fo user", party_uuid=party['party_uuid'])
```
andrewmil: Spelling; `fo` -> `for`

ras_party/controllers/account_controller.py
```diff
@@ -6,6 +6,7 @@
 from itsdangerous import SignatureExpired, BadSignature, BadData
 from requests import HTTPError
 from sqlalchemy import orm
+from werkzeug.exceptions import InternalServerError, BadRequest, NotFound, Conflict
```
andrewmil: Imports should be alphabetised

ras_party/controllers/account_controller.py
```diff
@@ -85,13 +88,13 @@ def post_respondent(party, tran, session):
     try:
         survey_id = collection_exercise['surveyId']
     except KeyError:
-        raise ClientError("There is no survey bound for this user", party_uuid=party['party_uuid'])
+        logger.debug("No survey bound fo user", party_uuid=party['party_uuid'])
+        raise BadRequest("There is no survey bound for this user")
```
andrewmil: I don't understand this check, if we successfully retrieve a collection exercise it will have a survey ID with it.
I think we can remove this whole bit of validation. 

If not then it should be InternalServerError instead

ras_party/controllers/account_controller.py
```diff
@@ -85,13 +88,13 @@ def post_respondent(party, tran, session):
...
 
     business = query_business_by_party_uuid(business_id, session)
     if not business:
-        raise ClientError("Could not locate business when creating business association",
-                          business_id=business_id,
-                          status=404)
+        logger.debug("Could not locate business when creating business association", business_id=business_id)
+        raise NotFound("Could not locate business when creating business association")
```
andrewmil: This is saying that the party is not found for the business_id from the case that we requested. If we get this error it feels like a InternalServerError instead of 404

ras_party/controllers/account_controller.py
```diff
@@ -567,9 +592,8 @@ def add_new_survey_for_respondent(payload, tran, session):
         """
         business = query_business_by_party_uuid(business_id, session)
         if not business:
-            raise ClientError("Could not locate business when creating business association",
-                              business_id=business_id,
-                              status=404)
+            logger.debug("Could not find business", business_id=business_id)
+            raise NotFound("Could not locate business when creating business association")
```
andrewmil: business_id comes from the case we retrieve here. If this doesn't correspond to a real business there's a problem with our data and I think this should be InternalServerError

ras_party/controllers/business_controller.py
```diff
@@ -1,15 +1,21 @@
 import uuid
 
+import logging
```
andrewmil, line 2: logging is in the standard library so should be at the top

ras_party/views/business_view.py
```diff
@@ -60,8 +65,14 @@ def get_business_by_ref(ref):
     verbose = request.args.get('verbose', '')
     verbose = True if verbose and verbose.lower() == 'true' else False
 
-    response = business_controller.get_business_by_ref(ref, verbose=verbose)
-    return jsonify(response)
+    business = business_controller.get_business_by_ref(ref, verbose=verbose)
+
+    if not business:
+        does_not_exist = "Business with reference does not exist"
+        logger.info(does_not_exist, reference=ref, url=request.url, status=404)
```
andrewmil: weird to define this string as a variable here first and to log out the url

ras_party/controllers/account_controller.py
andrewmil: Untested

ras_party/controllers/account_controller.py
```diff
@@ -126,13 +129,13 @@ def post_respondent(party, tran, session):
         _send_email_verification(respondent.party_uuid, party['emailAddress'])
     except (orm.exc.ObjectDeletedError, orm.exc.FlushError, orm.exc.StaleDataError, orm.exc.DetachedInstanceError):
         logger.error('Error updating database for user', party_id=party['id'])
-        raise RasError("Error updating database for user", status=500, party_id=party['id'])
+        raise InternalServerError(f"Error updating database for user {party['id']}")
```
andrewmil: All of these exceptions are untested

ras_party/controllers/account_controller.py
```diff
@@ -275,7 +286,9 @@ def change_respondent_password(token, payload, tran, session):
             password=new_password)
 
     if oauth_response.status_code != 201:
-        raise RasError("Failed to change respondent password")
+        logger.error("Unexpected response from auth service, unable to change user password",
```
andrewmil, line 274: Untested

ras_party/controllers/account_controller.py
```diff
@@ -465,7 +486,9 @@ def update_verified_email_address(respondent, tran, session):
         account_verified='true')
 
     if oauth_response.status_code != 201:
-        raise RasError("Failed to change respondent email", respondent_id=str(respondent.party_uuid))
+        logger.error("Unexpected response from auth service, unable to change user email",
```
andrewmil, line 474: Untested

andrewmil (changes requested):
Looks good, acceptance/unit tests pass

Left a few comments about which exceptions we should be throwing where
Also I think this is a good opportunity to make sure that these exceptions are being tested so probably want to get the test coverage up

ras_party/error_handlers.py
```diff
@@ -45,15 +27,15 @@ def http_error(error):
...
                      url=request.url,
                      errors=errors,
                      status=response.status_code)
     return response
 
 
 @blueprint.app_errorhandler(Exception)
-def exception_error(_):
+def exception_error(e):
     logger.exception('Uncaught exception', status=500)
```
andrewmil: I think we need to question whether this exception handler is giving us any value, right now it's just catching it and doing nothing with it.

shahi645:
Im happy with the changes mostly, just some test coverage needed to be added from @tej99 which andrew has picked up on, then good to get this in

ras_party/controllers/party_controller.py
```diff
@@ -58,16 +67,18 @@ def get_party_by_id(sample_unit_type, id, session):
     if sample_unit_type == Business.UNIT_TYPE:
         business = query_business_by_party_uuid(id, session)
         if not business:
-            raise ClientError("Business with id does not exist", business_id=id, status=404)
+            logger.debug("Business with id does not exist", business_id=id, status=404)
+            raise NotFound("Business with id does not exist")
         return business.to_party_dict()
     elif sample_unit_type == Respondent.UNIT_TYPE:
         respondent = query_respondent_by_party_uuid(id, session)
```
andrewmil, line 74: Untested

ras_party/controllers/party_controller.py
```diff
@@ -58,16 +67,18 @@ def get_party_by_id(sample_unit_type, id, session):
     if sample_unit_type == Business.UNIT_TYPE:
         business = query_business_by_party_uuid(id, session)
         if not business:
-            raise ClientError("Business with id does not exist", business_id=id, status=404)
+            logger.debug("Business with id does not exist", business_id=id, status=404)
```
andrewmil, line 70: Untested

andrewmil (approved):
There's still a couple of points that aren't tested but acceptance tests are passing and test coverage has been bumped up significantly so good to go I think :+1:

PR #177, opened 2018-09-26
Remove unused env var for ras api gateway

gemmai95:
# Motivation and Context

There are a number of environment variables that are no longer used. In party these are the environment variables for RAS API Gateway that no longer exists.

# What has changed

# How to test?

Ran tests and ran up application and checked works as expected with acceptance tests.

# Links

[Trello] https://trello.com/c/v6e0TLEJ


AndrewTorrance (approved):

shahi645 (approved):

tej99 (approved):
What even is the ras-api-gateway :/

PR #181, opened 2018-10-22
Handle list of enrolment statuses

gemmai95:
# Motivation and Context

Pending enrolments aren't currently appearing in the print file this should solve this.

# What has changed

Change the get party by sample unit type, survey and enrolment status to accept list.

# How to test?

Set up a business with only a pending enrolment change the action rule for the pending enrolments survey and make sure action is requesting the business details then respondent.

# Links

Trello: https://trello.com/c/Cwkh7tWn


andrewmil (commented):
Tested locally in docker, action requests now look like they're processed properly with pending enrolments

andrewmil (approved):

LJBabbage (approved):

PR #189, opened 2019-01-25
Indexes added via alembic script

AndrewTorrance:
# Motivation and Context
Party is running slow , on investigation some tables where poorly indexed.

# What has changed
Added indexes for business, business_respondent, enrolment, respondent, pending_enrolment, and business_attributes tables. Ran tests locally to see impact. However due to the small table size locally we suspect they will have more impact in preprod and prod.

# How to test?
Locally run the alembic script upgrade , verify new indexes added, then run downgrade . Make sure no errors

NOTE:
Before running this in pre-prod or prod check there is enough space. Locally these indexes are roughly 50% of the table size after the upgrade. 
Also be aware that we do not know how long they will take to run in prod so we should measure this in pre-prod

### DIFF
--- a/migrations/versions/fa52b1008d60_new_business_attributes_indexes.py
+++ b/migrations/versions/fa52b1008d60_new_business_attributes_indexes.py
@@ -1,0 +1,67 @@
+"""new business attributes indexes
+
+Revision ID: fa52b1008d60
+Revises: 613bf3a40948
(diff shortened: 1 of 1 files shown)


ryangrundy7 (approved):
LGTM 👍

NickGrantX (commented):
LGTM 👍

PR #191, opened 2019-02-15
751 index bug fix

AndrewTorrance:
# Motivation and Context
Changes made via Alembic not reaching prep-prod. This is due to alembic not able to cope with running from multiple scripts at the same time. (e.g starting 2 ras parties in preprod) 
It leaves the db in an indeterminate state 

# What has changed
Removed the running of alembic update scripts until fix is implemented. Created a SQL script for the specific change that was causing the issue which can be run manually during deploy. 
Also added indexes to model so that model agrees with expected state after running script

# How to test?
Difficult to test. Locally-
1) To test Model: Run the service and validate that the indexes are in the DB
2) To test script: Drop the partysvc schema. Comment out the Index creation lines in the model. Debug any test again breaking before end so that DB does not have indexes. Now run the script against the partysvc. And validate all indexes are created. Remember to reinstate the index creation in the models.

# How to Deploy

Down PartySvc
Get ready to time how long the index creation takes, just in case the time is significant.
Run sql script to install indexes. 
Validate indexes are created. 
Set the value in the alembic_version table to 613bf3a40948 to reflect the high water mark of the alembic chain
Deploy new party service. 

# Links
https://trello.com/c/MMdqguDK


NickGrantX:
I've tried this branch and the acceptance tests keep failing - 3 failures. Not sure if it's related or not though, because your changes looked ok, by eye.

I tested the index creation script, and that works OK.

run.py
```diff
@@ -82,8 +82,25 @@ def current_request():
         else:
             logger.info("Schema exists.", schema=db_schema)
 
-            logger.info("Running Alembic database upgrade")
-            command.upgrade(alembic_cfg, "head")
+            # START 15-Feb-2019 Bug Fix 751
```
rhiGriff: Why do we need to put in a log line saying that the alembic scripts are not being run if they have been removed from the code?

NickGrantX:
Acceptance tests keep failing - something to do with social and sure to be unrelated, but hard to verify that this PR hasn't broken anything at the moment. Pipeline's been broken for 4 days too.

run.py
AndrewTorrance: I ummed and arr'd about that one . Since it is a temporary fix I was minded to leave the comment there , which is not something I normally advocate. Will bring up with John in t morning

owencjones (approved):
LGTM

NickGrantX (approved):
The social ATs are very unreliable, and it failed every time for me locally, but those tests are not related to your PR so I'm approving.

Good work 👍

owencjones:
We've waiting long enough for Codacy to pull it's finger out here.

PR #192, opened 2019-02-22
215 respondent search

AndrewTorrance:
# Motivation and Context
Needed ability to search for respondents by a combination of first name , last name and email address , so that internal users can resolve respondents.
 
# What has changed

Added new end point that supports get respondents by firstName, lastName and email address. These are all case insensitive. first name and last name match the start of the names, email should be contained within the email address. Note the search supports the use of '%' as a wildcard for any number of characters.
The end point supports pagination in the form of page and limit. Page starting at 1 and limit being the maximum records per page. The results also include a total, this is the total number of records that match. So that if I have for example 1000 matching records but limit is 10 then total will be 1000. 
Updated API.MD
Added unit tests
created sql script
Modified model to create indexes when creating a new db, note sql script needs to get run for preprod and prod
Also removed calls to alembic due to earlier bug fix

# How to test?
Run the unit tests  

# How to deploy
Note this includes a sql script that should be run as part of deploy on preprod and prod


migrations/README
```diff
@@ -30,3 +30,7 @@ When another change has been made to the database and you require another versio
 ```
 alembic revision -m "create account table"
 ```
+
+### Use of ORM
+Sql Alchemy will quite happily modify the tables via  models.Base.metadata.create_all(engine) (in run.py)  in
+this project. So an alembic script
```
insacuri: So an alembic script... what?!  Don't leave me in suspense!

ras_party/views/respondent_view.py
```diff
@@ -31,18 +32,56 @@ def get_pw(username):
...
+    Validates the combination of parameters for get respondents
+    :param ids:
+    :param first_name:
+    :param last_name:
+    :param email:
+    """
+
+    if not (ids or first_name or last_name or email):
+        logger.debug("Invalid params: either id , first_name or last_name or email are required")
+        raise BadRequest(f"id , first_name or last_name or email are required")
```
insacuri: Super minor, but should be `f"id, first_name......`

insacuri: Also why is this an f string?

AndrewTorrance: Good catch, changing

migrations/README
AndrewTorrance: Good catch , use of alembic changed half way through , changing

insacuri (changes requested):
Other than the minor comments, this looks good I think.  Good job on updating the documentation on the api ^^

insacuri (approved):
LGTM

owencjones (approved):
Looks fine to me.  Like the restructuring of the endpoints, and the approach you've taken

PR #195, opened 2019-05-14
Added user claim validation endpoint

AndrewTorrance:
# Motivation and Context
Prior to this change if a valid user obtained a url for a conversation then they were able to take part in that conversation. This enables the closing of that vulnerability by providing an endpoint that can be called to validate that a user has a claim .

This is the second attempt at this, the first passed the entire set of claims in the JWT, but size became an issue there , so this method was used instead

# What has changed

New endpoint Get respondents/claim. This takes a user , survey and ru and returns either Valid or Invalid in the body depending on if the user has a valid claim
( This is more of an RPC style not RESTFUL, and the alternative was to hijack the status codes and use 200/403 for the same purpose , but since it is a GET claim then it should return 200 on success , but success does not indicate the status of the content)

# How to test?
This is covered by several unit tests that cover all the identified uses of this endpoint. Running these suffices.

# Links
https://trello.com/c/zqHOqb7O


ras_party/views/respondent_view.py
```diff
@@ -98,3 +98,27 @@ def change_respondent_details(respondent_id):
...
     return make_response('Successfully updated respondent details', 200)
+
+
+@respondent_view.route('/respondents/claim')
+def validate_respondent_claim():
+    """Validates if the respondent has a claim on a specific business and survey
+    Both valid and invalid claims return 200 with the difference being indicated in the response body.
+
+    Not RESTFUL style more like RPC, but the alternative is for duplication of business logic
+    in client services. There is an argument to use a 403 on invalid claims , but that should be a status on the
```
mfcrocker, line 109: Yeah, as discussed on the other PR the status codes are somewhat left wanting in this situation!

mfcrocker (approved):
Looks good, just be aware that the Codacy issue looks like a real one :)

ryangrundy7 (commented):
The default values for the auth service are set on the django one so it may be a good idea to set them to the new auth service, I created a account in the old and wondered why I couldn't log in 😆 This could be a separate card though 👍

ryangrundy7 (approved):
LGTM 👍 There are a lot of warnings when running the tests in this PR and SM, maybe we should have a card on the backlog to address them?

insacuri:
@ryangrundy7 The fixes to those warnings were already in master.  Just merged it, reran the tests and all the warnings are gone

PR #201, opened 2019-08-15
Added rollbacks and the use of sub transactions

AndrewTorrance:
# Motivation and Context
If auth registration failed on add new user then the records where left in the party service.
Added sub transactions so that roll back can be added on auth registration fail

# What has changed
removed @transaction , this was an app level transaction not db transaction and was not used
Added sub transactions so that we can roll back to specific points

# How to test?
run unit and acceptance tests . A difficult one to test manually.

# Links
https://trello.com/c/UtaGbEk1

### DIFF
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -106,8 +106,8 @@ def post_respondent(party, session):
 def _add_enrolment_and_auth(business, business_id, case_id, party, session, survey_id, translated_party):
     """Create and persist new party entities and attempt to register with auth service.
     Auth fails lead to party entities being rolled back.
+    The Contxet manager commits to session, If db fails after that and before main commit then db state is unknown
     """
-    session.begin_nested()  # Create a transaction SAVEPOINT
     try:
         with session.begin_nested():
 
@@ -143,7 +143,6 @@ def _add_enrolment_and_auth(business, business_id, case_id, party, session, surv
             session.rollback()  # Rollback to SAVEPOINT
             oauth_response.raise_for_status()
 
-        session.commit()  # Commits to sub transaction
         session.commit()  # Full session commit
 
     logger.info("New user has been registered via the auth-service")
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -33,13 +33,11 @@ NO_RESPONDENT_FOR_PARTY_ID = 'There is no respondent with that party ID '
 EMAIL_VERIFICATION_SENT = 'A new verification email has been sent'
 
 
-@transactional
 @with_db_session
-def post_respondent(party, tran, session):
+def post_respondent(party, session):
     """
     Register respondent and set up pending enrolment before account verification
     :param party: respondent to be created details
-    :param tran
     :param session
     :return: created respondent
     """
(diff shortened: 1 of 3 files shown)


mfcrocker (approved):

ras_party/controllers/account_controller.py
```diff
@@ -95,26 +93,63 @@ def post_respondent(party, tran, session):
...
+                                                 survey_id=survey_id)
+            Enrolment(business_respondent=br,
+                      survey_id=survey_id,
+                      status=EnrolmentStatus.PENDING)
+
+            session.add(respondent)
+            session.add(pending_enrolment)
+
+    except Exception as exception:
+        logger.error('Party service db post respondent caused exception', party_uuid=translated_party['party_uuid'],
```
insacuri: You can rewrite this bit to 
```
except Exception:
    logger.exception('Party service db post respondent caused exception', party_uuid=translated_party['party_uuid'])
    raise
```

I'd also say whilst it's fresh in your head, can we have more specific exceptions instead of the top level `Exception` so we have a better idea what can go wrong down the line.

insacuri (changes requested):

ras_party/controllers/account_controller.py
```diff
@@ -95,26 +93,62 @@ def post_respondent(party, tran, session):
...
+                     error=exception)
+        raise  # re raise the exception aimed at the generic handler
+    else:
+        # Register user to auth server after successful commit
+        oauth_response = OauthClient().create_account(party['emailAddress'], party['password'])
+        if not oauth_response.status_code == 201:
+            logger.info('Registering respondent auth service responded with', status=oauth_response.status_code,
```
necrophonic, line 141: Should probably be an `error` rather than an `info` as it's unexpected

ras_party/controllers/account_controller.py
necrophonic: I'd also write it maybe more like: "Exception occurred while attempting to enrol new user" - we'll know it's the party service et al from the metadata that'll be attached to the logs

ras_party/controllers/account_controller.py
```diff
@@ -95,26 +93,62 @@ def post_respondent(party, tran, session):
...
 
-    register_user(party, tran)
-    disable_iac(party['enrolmentCode'], case_id)
     return respondent.to_respondent_dict()
 
 
+def _add_enrolment_and_auth(business, business_id, case_id, party, session, survey_id, translated_party):
+    """Create and persist new party entities and attempt to register with auth service.
+    Auth fails lead to party entities being rolled back.
+    The Contxet manager commits to session, If db fails after that and before main commit then db state is unknown
```
insacuri: Typo in comment

ras_party/controllers/account_controller.py
```diff
@@ -95,26 +93,62 @@ def post_respondent(party, tran, session):
...
+                                                 business_id=business_id,
+                                                 survey_id=survey_id)
+            Enrolment(business_respondent=br,
+                      survey_id=survey_id,
+                      status=EnrolmentStatus.PENDING)
+
+            session.add(respondent)
+            session.add(pending_enrolment)
+
+    except Exception as exception:
```
insacuri: Can we have this work for all the specific exceptions instead of just the generic exception?

ras_party/controllers/account_controller.py
```diff
@@ -95,26 +93,62 @@ def post_respondent(party, tran, session):
...
+        # Register user to auth server after successful commit
+        oauth_response = OauthClient().create_account(party['emailAddress'], party['password'])
+        if not oauth_response.status_code == 201:
+            logger.info('Registering respondent auth service responded with', status=oauth_response.status_code,
+                        content=oauth_response.content)
+
+            session.rollback()  # Rollback to SAVEPOINT
+            oauth_response.raise_for_status()
+
+        session.commit()  # Full session commit
```
insacuri, line 147: Can we get a comment either here or at the top that acknowledges that if this commit fails then we're going to be in a half complete state that we'll just have to handle on the rare case by case basis.
Either that or handle this unlikely (but possible) issue?

insacuri (changes requested):

insacuri (approved):

PR #206, opened 2019-10-01
Delete user

insacuri:
# Motivation and Context
We need the ability to delete respondents as part of GDPR requests and general housekeeping.  We've made an endpoint to delete users instead of a script as it's much safer!

# What has changed
  - Delete endpoint created that will remove a respondent based on their party_uuid
  - Tox removed from testing process, replaced with pytest (that it was running anyway).  Matches how we test in ras-frontstage
  - Linted and tidied up API.md as well updating with new endpoint

# How to test?
- Run acceptance tests to set up seed data
- Go into database and find a party_uuid in partysvc -> respondents table.  Make a note of the `id` of the user too.
- Hit the endpoint with that uuid.  You should notice the following things have been deleted:
  - Business respondent with id should be deleted
  - Pending enrolment with id should be deleted (if it existed)
  - Enrolment with id should be deleted
  - Respondent with party_uuid should be deleted.

Also test it by hitting it with a non-uuid, and a user that doesn't exist, you should get the appropriate error messages

# Links
https://trello.com/c/ldqHvnFq


mfcrocker (approved):

jins-ons (approved):
Changes Looks fine

PR #208, opened 2019-10-16
Replace delete by id with delete by email

insacuri:
# Motivation and Context
The original delete respodent in party did it by ID.  This is tricky because the auth service can only do it by email.  By doing the delete by email, the script to actually delete a user becomes much simpler as we only need 1 thing (the users email) rather then having to take an email and resolve it to a party_uuid before being able to do anything.

# What has changed
Added a DELETE /respondents/email endpoint
Removed the DELETE /respondents/id/uuid endpoint

# How to test?
- Seed the party database with the acceptance tests.
- Find a user to delete
- Hit the endpoint with that email.  It should fail.

- Additionally, try and hit it with empty json, json with a blank email address and with content that isn't json.  All of these should throw 400 errors.

# Links
https://trello.com/c/GzZO9UbF

### DIFF
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -2,14 +2,16 @@ import logging
 import uuid
 
 import structlog
-from werkzeug.exceptions import BadRequest, NotFound
+from werkzeug.exceptions import BadRequest, NotFound, UnprocessableEntity
+from sqlalchemy.orm.exc import NoResultFound, MultipleResultsFound
 
 from ras_party.controllers.account_controller import change_respondent
 from ras_party.models.models import Enrolment, BusinessRespondent, PendingEnrolment, Respondent
 from ras_party.controllers.queries import query_respondent_by_party_uuid, \
     query_respondent_by_email, update_respondent_details, query_respondent_by_names_and_emails, \
-    query_respondent_by_party_uuids
+    query_respondent_by_party_uuids, query_single_respondent_by_email
 from ras_party.support.session_decorator import with_db_session
+from ras_party.support.util import obfuscate_email
 
 
 logger = structlog.wrap_logger(logging.getLogger(__name__))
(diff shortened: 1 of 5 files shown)


ras_party/controllers/respondent_controller.py
```diff
@@ -70,35 +72,37 @@ def get_respondent_by_id(respondent_id, session):
...
+        respondent = query_single_respondent_by_email(email, session)
+    except NoResultFound:
+        logger.error("Respondent with email does not exist", email=obfuscate_email(email))
+        raise NotFound("Respondent with email does not exist")
+    except MultipleResultsFound:
+        logger.error("Multiple respondents found for email", email=obfuscate_email(email))
+        raise UnprocessableEntity("Multiple users found, unable to proceed")
 
-    logger.info("Found respondent", party_uuid=str(respondent.party_uuid), id=respondent.id)
+    logger.info("Found respondent", email=respondent.email_address, party_uuid=respondent.party_uuid, id=respondent.id)
```
rhiGriff: This log line either cannot log the email or the email needs to be obfuscated

rhiGriff (changes requested):

ras_party/controllers/respondent_controller.py
insacuri: Good catch!  Thought I'd got them all

ras_party/controllers/respondent_controller.py
```diff
@@ -70,35 +72,40 @@ def get_respondent_by_id(respondent_id, session):
 
 
 @with_db_session
-def delete_respondent_by_id(party_uuid, session):
+def delete_respondent_by_email(email, session):
     """
-    Delete a Respondent by its Party ID
+    Delete a Respondent by its email
```
AndrewTorrance, line 77: Doc string needs updating to be email

rhiGriff (approved):

ras_party/views/respondent_view.py
```diff
@@ -99,6 +93,22 @@ def get_respondent_by_email():
     return jsonify(response)
 
 
+@respondent_view.route('/respondents/email', methods=['DELETE'])
+def delete_respondent_by_id():
```
AndrewTorrance: minor :- given the mismatch between use of id in the name and email it may be worth a comment ?

AndrewTorrance (changes requested):
doc string comments

ras_party/controllers/respondent_controller.py
insacuri: Updated it already!

ras_party/views/respondent_view.py
insacuri: Done

insacuri:
There's a missing unit test for MultipleResultsFound.  This needs to be done too and I was just being lazy by not writing it sorry >_<

AndrewTorrance (approved):

PR #212, opened 2019-10-31
Modify party search

insacuri:
# Motivation and Context
The search for business information is very slow, especially when few or no search terms are included in the search. 

# What has changed
- name and trading as columns added to business_attributes table
- scripts added to add new columns and to put example data into the database
- search_business function modified to be a bit smarter about searching for a full ru_ref while making every other type of search use the column data instead of faster.

# How to test?
- Delete docker image for postgres
- rebuild docker containers, look at the partysvc -> business_attributes table.  The name and trading_as columns should be added (these are generated from the model when the party schema is missing)

- Next, seed some data with the acceptance tests setup.
- sign into response-operations-ui and search for a reporting unit.  Everything should still work.

- Finally (to test the speed), go to the scripts directory, and `pipenv run python insert_into_party.py`.  By default this will add 20k businesses and 100k rows in to the business attributes table but you can change the numbers.  Do this a few times then do an empty search, a search with a few characters and an ru_ref search.  These should all be reasonably fast

# Links
https://trello.com/c/WerjgQnt


insacuri:
Just added don't merge label as the data setup in acceptance tests is broken somehow?

insacuri:
Removed the label as we fixed the issue (sample service was posting top party with a null name, which used to be fine, but wasn't with a change to the schema.  We've fixed this now).

I also verified that the name and trading as columns are being populated as expected

ras_party/controllers/party_controller.py
```diff
@@ -21,8 +21,10 @@ def parties_post(party_data, session):
     :param party_data: packet containing the data to post
     :type party_data: JSON data maching the schema described in schemas/party_schema.json
     """
+    logger.info("party data in controller", party_data=party_data)
```
mfcrocker: Does this have personal data in it?

ras_party/controllers/party_controller.py
```diff
@@ -21,8 +21,10 @@ def parties_post(party_data, session):
...
     errors = Business.validate(party_data, current_app.config['PARTY_SCHEMA'])
     if errors:
+        logger.info("errors", errors=errors)
```
mfcrocker: This seems near-identical to the line below - we should have one or the other

mfcrocker (changes requested):
Couple of questions around logging else LGTM

ras_party/controllers/party_controller.py
necrophonic: Agree. Feels like debugging code that's been left in

ras_party/controllers/party_controller.py
necrophonic: Overhang of debugging code?

ras_party/controllers/queries.py
```diff
@@ -174,22 +174,30 @@ def search_businesses(search_query, session):
     :return: list of businesses
     """
     logger.info('Searching businesses by name with search query', search_query=search_query)
-    filters = list()
-    name_filters = list()
-    trading_as_filters = list()
+    if len(search_query) == 11 and search_query.isdigit():
+        logger.info("Query looks like an ru_ref, searching only on ru_ref", search_query=search_query)
```
necrophonic: Is it useful to keep logging the query? Should we just tie this together with a context bound to the logger?

scripts/insert_into_party.py
```diff
@@ -0,0 +1,58 @@
+import sys
```
necrophonic: Scripts need better naming - "insert_into_party" doesn't give any context of _what_ it inserts and _why_. If it's just for testing it should be clearly separated from scripts used for deployment/rollback.

scripts/insert_into_party.py
```diff
@@ -0,0 +1,58 @@
...
+    connection.commit()
+    count = cursor.rowcount
+    print (count, "Record inserted successfully into mobile table")
+
+except (Exception, psycopg2.Error) as error:
+    if(connection):
+        print("Failed to insert record into mobile table", error)
+
+finally:
+    #closing database connection.
```
necrophonic: Redundant comment

scripts/insert_into_party.py
```diff
@@ -0,0 +1,58 @@
...
+        cursor.execute(fourthba_insert_query)
+        cursor.execute(fifthba_insert_query)
+
+    end = datetime.now()
+    print(f"started at: {start}")
+    print(f"ended at: {end}")
+
+    connection.commit()
+    count = cursor.rowcount
+    print (count, "Record inserted successfully into mobile table")
```
necrophonic: I possibly don't have the context, but why does it say _mobile table_?

scripts/insert_into_party.py
```diff
@@ -0,0 +1,58 @@
...
+        thirdba_insert_query = f"""INSERT INTO partysvc.business_attributes(business_id, sample_summary_id, collection_exercise, "attributes", created_on, "name", trading_as) VALUES('{business_id}', '{uuid.uuid4()}', '{uuid.uuid4()}', '{{"name": "{name}", "trading_as": "{trading_as}"}}', '{datetime.now()}', '{name}', '{trading_as}');"""
+        fourthba_insert_query = f"""INSERT INTO partysvc.business_attributes(business_id, sample_summary_id, collection_exercise, "attributes", created_on, "name", trading_as) VALUES('{business_id}', '{uuid.uuid4()}', '{uuid.uuid4()}', '{{"name": "{name}", "trading_as": "{trading_as}"}}', '{datetime.now()}', '{name}', '{trading_as}');"""
+        fifthba_insert_query = f"""INSERT INTO partysvc.business_attributes(business_id, sample_summary_id, collection_exercise, "attributes", created_on, "name", trading_as) VALUES('{business_id}', '{uuid.uuid4()}', '{uuid.uuid4()}', '{{"name": "{name}", "trading_as": "{trading_as}"}}', '{datetime.now()}', '{name}', '{trading_as}');"""
+        print(firstba_insert_query)
```
necrophonic: Why printing first query to the terminal?

scripts/insert_into_party.py
```diff
@@ -0,0 +1,58 @@
...
+        thirdba_insert_query = f"""INSERT INTO partysvc.business_attributes(business_id, sample_summary_id, collection_exercise, "attributes", created_on, "name", trading_as) VALUES('{business_id}', '{uuid.uuid4()}', '{uuid.uuid4()}', '{{"name": "{name}", "trading_as": "{trading_as}"}}', '{datetime.now()}', '{name}', '{trading_as}');"""
+        fourthba_insert_query = f"""INSERT INTO partysvc.business_attributes(business_id, sample_summary_id, collection_exercise, "attributes", created_on, "name", trading_as) VALUES('{business_id}', '{uuid.uuid4()}', '{uuid.uuid4()}', '{{"name": "{name}", "trading_as": "{trading_as}"}}', '{datetime.now()}', '{name}', '{trading_as}');"""
+        fifthba_insert_query = f"""INSERT INTO partysvc.business_attributes(business_id, sample_summary_id, collection_exercise, "attributes", created_on, "name", trading_as) VALUES('{business_id}', '{uuid.uuid4()}', '{uuid.uuid4()}', '{{"name": "{name}", "trading_as": "{trading_as}"}}', '{datetime.now()}', '{name}', '{trading_as}');"""
```
necrophonic: Could this just be a loop?

necrophonic (changes requested):
Given deploying these changes involves manual steps outside of tag-and-build there needs to be a deployment plan (and rollback steps)

ras_party/controllers/party_controller.py
insacuri: It was yeah.  Sorry for wasting your guys time with dumb stuff in the PR >_<

ras_party/controllers/queries.py
insacuri: Agreed, created a bound logger with that field bound to it.

scripts/insert_into_party.py
insacuri: Yup

scripts/insert_into_party.py
insacuri: The context you're missing is the code I copy/pasta'd from the internet ^^ Removed it

scripts/insert_into_party.py
insacuri: Debugging, tidied it up now

scripts/insert_into_party.py
insacuri: It's now a loop

scripts/insert_into_party.py
insacuri: Renamed to create_example_business_data.  I'll put an explanation of what the script does in a docstring as moving it into a deeper folder would mean having to mess around with paths some more which I don't really want to do (as the first 4 lines of this are messy enough as it is)

ras_party/controllers/queries.py
```diff
@@ -173,23 +173,33 @@ def search_businesses(search_query, session):
...
-    trading_as_filters = list()
+    bound_logger = logger.bind(search_query=search_query)
+    bound_logger.info('Searching businesses by name with search query')
+    if len(search_query) == 11 and search_query.isdigit():
+        bound_logger.info("Query looks like an ru_ref, searching only on ru_ref")
+        result = session.query(BusinessAttributes.name, BusinessAttributes.trading_as, Business.business_ref)\
+            .join(Business).filter(Business.business_ref == search_query).distinct().all()
+        if result:
+            return result
+        bound_logger.info("Didn't find an ru_ref, seraching everything")
```
necrophonic: :art: Typo seraching > searching

ras_party/controllers/queries.py
```diff
@@ -173,23 +173,33 @@ def search_businesses(search_query, session):
     :param search_query: the search query
     :return: list of businesses
     """
```
necrophonic, line 175: May be worth some more explanation of what sort of things this function is expecting (what do we mean by "search_query"?) and the logic flow

mfcrocker (approved):

ras_party/controllers/queries.py
AndrewTorrance: Done

ras_party/controllers/queries.py
AndrewTorrance: Done

necrophonic (approved):

PR #219, opened 2019-11-14
Search pagination

AndrewTorrance:
# Motivation and Context
When the search terms for ru's are vague it can attempt to return huge amounts of data . This pagination is aimed at stopping that causing errors 

# What has changed
Added limit and page params to the search and also added a total record count to the returned data ( needed for pagination) 

# How to test?
option 1
Run acceptance tests to populate data
Run  party and response ops on new branch. Search for '' and get paginated results 
option 2
run create_example_buisness_data after editing it for a large number of records ( say 100000).
search for '' and validate paginated return (In this case if you view details you will get an error) 

Also run tests (unit tests check for pagination controls). No acceptance tests added as covered in unit tests. 

# Links
https://trello.com/c/fqC2NEdK


mfcrocker (approved):
Looks good

ras_party/controllers/party_controller.py
```diff
@@ -21,8 +21,10 @@ def parties_post(party_data, session):
     :param party_data: packet containing the data to post
     :type party_data: JSON data maching the schema described in schemas/party_schema.json
     """
+    logger.info("party data in controller", party_data=party_data)
     errors = Business.validate(party_data, current_app.config['PARTY_SCHEMA'])
     if errors:
+        logger.info("errors", errors=errors)
```
insacuri: Debug logging?

ras_party/controllers/queries.py
```diff
@@ -199,10 +201,19 @@ def search_businesses(search_query, session):
...
         .join(Business)\
         .filter(and_(or_(*filters), BusinessAttributes.collection_exercise.isnot(None)))\
-        .distinct().all()
+        .distinct().order_by(BusinessAttributes.name)          # Build the query
+
+    results = query.limit(limit).offset((page-1)*limit).all()  # Execute the query
+    if page == 1 and len(results) < limit:
+        total_business_count = len(results)
+    else:
+        count_q = query.statement.with_only_columns([func.count(distinct(Business.business_ref))]).order_by(None)
```
insacuri, line 213: nice job with the only columns, I imagine it makes the search more efficient.
Do you need the order_by(None)?

ras_party/views/business_view.py
```diff
@@ -86,6 +86,8 @@ def put_business_attributes_ce(sample):
 @business_view.route('/businesses/search', methods=['GET'])
 def get_party_by_search():
     query = request.args.get('query', '')
+    page = int(request.args.get('page', default=1))
+    limit = int(request.args.get('limit', default=100))
 
-    response = business_controller.get_businesses_by_search_query(query)
-    return jsonify(response)
+    businesses, total_business_count = business_controller.get_businesses_by_search_query(query, page, limit)
+    return {'businesses': businesses, 'total_business_count': total_business_count}
```
insacuri: You need to jsonify the return value I think.  I believe it does some stuff for you under the hood like setting the mimetypes and status codes

insacuri (changes requested):
Just a few minor comments, but it's looking good from a code standpoint

ras_party/controllers/queries.py
AndrewTorrance: I think it could be left off but I think its an explicit instruction to the DBM  to ignore the sort column and hence gives it the option of selecting the most efficient way of getting the result without worrying about sort order .

ras_party/controllers/party_controller.py
AndrewTorrance: Not sure , I think you added according to my ui which is odd

ras_party/views/business_view.py
AndrewTorrance: Good catch thanks

ras_party/controllers/party_controller.py
AndrewTorrance: Removing

insacuri (approved):

PR #222, opened 2019-11-20
added disable enrolments by email

AndrewTorrance:
# Motivation and Context
There is a need to remove enrolments from a user months ahead of deleting them

# What has changed
Added new endpoint to disable all enrolments for a user based on email address

# How to test?
Run unit tests. Makes use of existing functionality to disable user , so no need to additionally test that specific functionality 

# Links
 https://trello.com/c/qRJOkASH


API.md
```diff
@@ -1,4 +1,4 @@
-# Party Service API
+ # Party Service API
```
insacuri: Why the space?

API.md
```diff
@@ -997,14 +997,23 @@ total would be 27
 }
 ```
 
-### Delete Respondent by ID
+### Delete Respondent by Email
 
-* `DELETE /party-api/v1/respondents/id/<id>`
-  * Deletes a respondent by their party_uuid.
+* `DELETE /party-api/v1/respondents/email`
+  * Deletes a respondent by their email address.
```
insacuri: Trailing whitespace

API.md
```diff
@@ -997,14 +997,23 @@ total would be 27
...
 
+### Disable all user enrolments by email
+
+* `PUT /party-api/v1/respondents/disable-user-enrolments`
+  * Disables all active enrolments for a respondent based on their email address. 
+  * Returns a 204 on success, a 400 if the id isn't a valid uuid and 404 if the user cannot be found.
+
+---
+
+
```
insacuri: Linter complaining about multiple lines (should only be 1)

API.md
```diff
@@ -997,14 +997,23 @@ total would be 27
...
+* `DELETE /party-api/v1/respondents/email`
+  * Deletes a respondent by their email address. 
   * Returns a 204 on success, a 400 if the id isn't a valid uuid and 404 if the user cannot be found.
 
 ---
 
+### Disable all user enrolments by email
+
+* `PUT /party-api/v1/respondents/disable-user-enrolments`
+  * Disables all active enrolments for a respondent based on their email address.
```
insacuri: Trailing whitespace

ras_party/controllers/account_controller.py
```diff
@@ -153,31 +154,43 @@ def _add_enrolment_and_auth(business, business_id, case_id, party, session, surv
...
+                                               business_id=payload['business_id'],
+                                               status=payload['change_flag'],
+                                               session=session)
+
+
+def _change_respondent_enrolment_status(respondent, survey_id, business_id, status, session):
+
     logger.info("Attempting to change respondent enrolment",
-                respondent_id=respondent_id,
+                respondent_id=respondent.party_uuid,
```
insacuri, line 180: Does this change definitely not mess up any other code?  I'm asking incase another bit of code calls this function with respondent_id being the uuid

ras_party/controllers/account_controller.py
```diff
@@ -190,6 +203,37 @@ def change_respondent_enrolment_status(payload, session):
...
+    """Disables all enrolments for a respondent ,  returns the count of the removed enrolments"""
+
+    obfuscated_email = obfuscate_email(respondent_email)
+
+    logger.info('Disabling all enrolments for respondent', email=obfuscated_email)
+
+    removed_enrolments_count = 0
+
+    # raises errors if none or multiple, unusual import to avoid circular references
+    respondent = __import__('ras_party').controllers.respondent_controller.get_single_respondent_by_email(
```
insacuri: You'll have to explain this to me because this looks wrong even with the comment explaining why it's not wrong ^^

insacuri (changes requested):

ras_party/controllers/respondent_controller.py
```diff
@@ -74,7 +74,7 @@ def get_respondent_by_id(respondent_id, session):
 @with_db_session
 def delete_respondent_by_email(email, session):
     """
-    Delete a Respondent by its email
+    Delete a Respondent by its email fter disabling all their enrolments
```
insacuri: Spelling mistake

ras_party/views/account_view.py
```diff
@@ -113,6 +113,21 @@ def change_respondent_enrolment_status():
     return make_response(jsonify('OK'), 200)
 
 
+@account_view.route('/respondents/disable-user-enrolments', methods=['PUT'])
```
insacuri: I think this should be a POST.  A put usually contains everything that is going to replace what is there.  Here, we're just providing an email and saying 'go change it for us'

ras_party/controllers/respondent_controller.py
insacuri: Also, this isn't true.  This function doesn't care about the state of enrolments.  All this function does is delete a respondent.

ras_party/controllers/account_controller.py
```diff
@@ -153,31 +154,43 @@ def _add_enrolment_and_auth(business, business_id, case_id, party, session, surv
 @with_db_session
 def change_respondent_enrolment_status(payload, session):
     """
-    Change respondent enrolment status for business and survey
-    :param payload:
-    :param session:
-    :return:
+    Change respondent enrolment status for business and survey, takes params from a payload dict
```
insacuri:
```suggestion
    Change respondent enrolment status for business and survey. Takes params from a payload dict
```

ras_party/controllers/account_controller.py
insacuri: We've talked and are going to move some code from the respondent controller into the account controller so we can remove this circular ref

ras_party/controllers/account_controller.py
```diff
@@ -190,6 +203,37 @@ def change_respondent_enrolment_status(payload, session):
...
+
+    for enrolment in enrolments:
+        _change_respondent_enrolment_status(respondent=respondent,
+                                            survey_id=enrolment.survey_id,
+                                            business_id=enrolment.business_id,
+                                            status='DISABLED',
+                                            session=session)
+        removed_enrolments_count += 1
+
+    # If no enrolments then logged count is zero
```
insacuri: No need for this comment, as you've defined the enrolment count slightly higher in the code

test/test_data/default_test_values.py
```diff
@@ -1,3 +1,4 @@
 DEFAULT_BUSINESS_UUID = '3b136c4b-7a14-4904-9e01-13364dd7b972'
 DEFAULT_SURVEY_UUID = 'cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87'
+ALTERNATE_SURVEY_UUID1 = '110711c3-0ac8-41d3-ae0e-567e5ea1ef87'
```
insacuri: Is there already a ALTERNATE_SURVEY_UUID?

insacuri (changes requested):

ras_party/views/account_view.py
```diff
@@ -113,6 +113,21 @@ def change_respondent_enrolment_status():
...
+    try:
+        email = request.get_json()['email']
+    except TypeError:
+        raise BadRequest('JSON payload not provided')
+    except KeyError:
+        raise BadRequest("Email key must be provided in the JSON payload")
+
+    removed_enrolment_count = account_controller.disable_all_respondent_enrolments(email)
+
+    return make_response(jsonify(f'{removed_enrolment_count} enrolments removed'), 200)
```
insacuri:
```suggestion
    return make_response(jsonify({"message": f'{removed_enrolment_count} enrolments removed'}), 200)
```

Responses should be in json format

API.md
AndrewTorrance: Must be fat fingers

ras_party/controllers/account_controller.py
AndrewTorrance: Only 2 calls made here , and in both party_uuid used .

ras_party/controllers/respondent_controller.py
AndrewTorrance: Missed it after reverting change , good catch

ras_party/views/account_view.py
AndrewTorrance: Disagree , if anything it should be a PATCH

test/test_data/default_test_values.py
AndrewTorrance: Nope , but thee was a 2 and I forgot to rename this when I backed it out, good catch

insacuri (approved):
LGTM

ras_party/views/account_view.py
```diff
@@ -113,6 +113,21 @@ def change_respondent_enrolment_status():
     return make_response(jsonify('OK'), 200)
 
 
+@account_view.route('/respondents/disable-user-enrolments', methods=['PATCH'])
```
mfcrocker, line 116: I think `PATCH` is technically correct but I'm a little worried that it might cause people some issues because no-one ever uses it :D

mfcrocker (approved):

PR #227, opened 2020-01-07
Remove frontstage through party

lauraedwards1:
# Motivation and Context

All the email communications on Frontstage needed for things like reset password etc went through Party, including for things as simple as verification link generation/checking.

# What has changed

# How to test?

# Links


mfcrocker (approved):

insacuri (approved):
LGTM,  I've not tested it yet but I'll approve the frontstage PR once I've tested/reviewed it

PR #230, opened 2020-01-28
Remove email from logging

jdfox:
PR
# Motivation and Context

From Technical Debt ticket "Remove email and name logging from party (2)", change was needed to remove excessive user detail from logs

# What has changed

In util.py, added handling of short, malformed and missing addresses.

No refactoring was required.

Unit test added to test_util.py 
# How to test?

make test
# Links

https://trello.com/c/vWfVa8OK/1463-remove-email-and-name-logging-from-party-2

### DIFF
--- a/ras_party/support/util.py
+++ b/ras_party/support/util.py
@@ -1,4 +1,8 @@
 import functools
+import structlog
+import logging
+
+logger = structlog.wrap_logger(logging.getLogger(__name__))
 
 
 def filter_dict(d, cb):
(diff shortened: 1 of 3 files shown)


test/test_util.py
```diff
@@ -18,3 +18,23 @@ def test_flatten_keys_produces_expected_results(self):
...
+        testAddresses = {
+            "example@example.com": "e*****e@e*********m",
+            "prefix@domain.co.uk": "p****x@d**********k",
+            "first.name@place.gov.uk": "f********e@p**********k",
+            "me+addition@gmail.com": "m*********n@g*******m",
+            "a.b.c.someone@example.com": "a***********e@e*********m",
+            "john.smith123456@londinium.ac.co.uk": "j**************6@l****************k",
+            "me!?@example.com": "m**?@e*********m",
+            "m@m.com": "m@m***m",
+            "joe.bloggs": "j********s",
```
insacuri, line 34: Are these last 3 cases why you changed the function?  If so, it's not really necessary as valid emails are <thing>@<domain>.<tld>
If that is the case, could you change the function back to the simpler version?

ras_party/support/util.py
```diff
@@ -64,7 +68,20 @@ def obfuscate_email(email):
...
+            prefix = f'{m[0][0]}{"*"*(len(m[0])-2)}{m[0][-1]}'
+        else:
+            prefix = m[0]
+        # If the domain is missing or 1 character, then we can't obfuscate it
+        if len(m) > 1 and len(m[1]) > 1:
+            domain = f'{m[1][0]}{"*"*(len(m[1])-2)}{m[1][-1]}'
+            return f'{prefix}@{domain}'
+        else:
+            return f'{prefix}'
+    except AssertionError:
```
insacuri: How does the assertionError get raised?

insacuri (changes requested):

test/test_util.py
jdfox: I added that last one ("joe.bloggs") to test_util.py, for completeness, but when we spun up dev it turned out that malformed addresses (no @domain etc) are rejected by the UI, so probably not needed. Happy to trim that test list. 
But most of the extended checking was added in order to allow obfuscation on the existing tests for malformed addresses in ras-party/test/**test_respondent_controller.py** to pass in a "make test", e.g. lines 206, 211, etc. Want me to chop those too, or is that out of scope here?

ras_party/support/util.py
jdfox: That was also raised by test_respondent_controller.py (see above), so that may also be overkill.

jdfox: I've replaced the try-except with a simpler set of ifs.

insacuri (approved):

warrenbailey (approved):

PR #243, opened 2020-03-19
Add DNS entries to Helm config

mfcrocker:
# Motivation and Context
Adds kube_dns addresses for all connecting services to allow us to have circular references


jdfox (approved):

jterryons (changes requested):
Frontstage is a special case, we'll have to leave that bit out for now

jterryons (approved):

PR #257, opened 2020-04-22
updated the logging in the notify controller

bkhan2:
# Motivation and Context
updating logging when notify notification doesn't get sent to be clearer

# What has changed
added URL and Status Code logging.

### DIFF
--- a/ras_party/controllers/notify_gateway.py
+++ b/ras_party/controllers/notify_gateway.py
@@ -51,7 +51,8 @@ class NotifyGateway:
             logger.info('Notification id sent via Notify-Gateway to GOV.UK Notify.', id=response.json()["id"])
         else:
             ref = reference if reference else 'reference_unknown'
-            raise RasNotifyError(f"There was a problem sending a notification to Notify-Gateway to GOV.UK Notify. URL = {url}, STATUS CODE = {status_code}",
+            raise RasNotifyError("There was a problem sending a notification to Notify-Gateway to GOV.UK Notify."
+                                 f" URL = {url}, STATUS CODE = {status_code}, MESSAGE = {response.text}",
                                  reference=ref)
 
     def request_to_notify(self, email, template_name, personalisation=None, reference=None):
--- a/ras_party/controllers/notify_gateway.py
+++ b/ras_party/controllers/notify_gateway.py
@@ -35,25 +35,24 @@ class NotifyGateway:
             logger.info("Notification not sent. Notify is disabled.")
             return
 
-        try:
-            notification = {
-                "emailAddress": email,
-            }
-            if personalisation:
-                notification.update({"personalisation": personalisation})
-            if reference:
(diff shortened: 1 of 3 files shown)


ras_party/controllers/notify_gateway.py
```diff
@@ -35,25 +35,25 @@ def _send_message(self, email, template_id, personalisation=None, reference=None
...
+
+        url = urlparse.urljoin(self.notify_url, str(template_id))
+        response = Requests.post(url, json=notification)
+        status_code = response.status_code
+
+        if status_code == 201:
             logger.info('Notification id sent via Notify-Gateway to GOV.UK Notify.', id=response.json()["id"])
-
-        except Exception as e:
+        else:
```
jterryons: is the deleted exception never thrown? we're only checking here for a 201. but codes 200 - 399 are valid and shouldnt raise an error. i think theres a method that might be useful response.raise_for_status() or something similar?

jterryons (changes requested):

ras_party/controllers/notify_gateway.py
mfcrocker: I disagree with this - 201 is the only valid non-error response from this endpoint and if we're considering the APIs a contract we should be willing to code that in

ras_party/controllers/notify_gateway.py
```diff
@@ -35,25 +35,25 @@ def _send_message(self, email, template_id, personalisation=None, reference=None
...
+
+        if 201 <= status_code <= 399:
             logger.info('Notification id sent via Notify-Gateway to GOV.UK Notify.', id=response.json()["id"])
-
-        except Exception as e:
+        else:
             ref = reference if reference else 'reference_unknown'
-            raise RasNotifyError("There was a problem sending a notification to Notify-Gateway to GOV.UK Notify",
-                                 error=e, reference=ref)
+            raise RasNotifyError("There was a problem sending a notification to Notify-Gateway to GOV.UK Notify."
```
mfcrocker: You don't need to pass these variables in as fstrings, `RasNotifyError` takes kwargs
```suggestion
            raise RasNotifyError("There was a problem sending a notification to Notify-Gateway.", url=url, status_code=status_code, message=response.text, reference=ref)
```

mfcrocker (changes requested):
Same comment around fstrings

ras_party/controllers/notify_gateway.py
jterryons: a 200/300 status code here will still have created the resource in whatever service is being called and will presumably be in a valid state. this error will not roll back any successful transactions and will be a nightmare to investigate any ops issue as a result of the error thrown. If a 201 is the only valid response then there should be no need to check it and a raise_for_status should suffice.

jterryons:
Similar to the frontstage change, this isnt worth the test effort.

PR #274, opened 2020-09-01
Make party config consistent with csv-worker

mfcrocker:
# Motivation and Context
Brings config for GCP config in line with the config used in csv-worker


jterryons (approved):

jterryons:
fix lint (or remove it completely) and it should be fine

PR #276, opened 2020-09-17
User data deletion for GDPR

creativeappdevamit:
## Objective of the PR
The Motivation for this PR is to provide a way for the user data to be deleted when triggered

Changes:
 * A new column `mark_for_deletion` (boolean/autogenerated) has been added to the responded table.
 * Old endpoints has modified to update the respondent table flag (`mark_for_deletion`) based on respondent's email address
 * New endpoint has been provided for the scheduler to trigger deletion of respondents marked for deletion.
 * New endpoint has been provided to batch requests

## How to test?
Smoke test


ras_party/batch_process/batch_request.py
```diff
@@ -0,0 +1,87 @@
+import json
```
mfcrocker: I don't think this should be in `/batch_process` when all the other endpoints are in `/views` - lets just have one place for routes to live

ras_party/batch_process/batch_request.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Endpoint Exposed for Kubernetes Cronjob to delete all respondents and
+    its associated data marked for deletion
+    """
+    respondent_controller.delete_respondents_marked_for_deletion()
+    return '', 204
+
+
+@batch_request.route('/batch/requests', methods=['POST'])
+def batch():
```
mfcrocker: Is this particularly performant? It'll obviously cut down on network traffic and is very flexible, but for the example JSON its still N SQL transactions where it could just be one update transaction.

If we're likely to want to do a mix of different requests in a batch then it should stay like this. If we're only ever going to want to batch patch respondents, we should write an endpoint to do just that and optimise DB transactions for it.

ras_party/views/respondent_view.py
```diff
@@ -91,6 +90,22 @@ def get_respondent_by_email():
     return jsonify(response)
 
 
+@respondent_view.route('/respondents/email', methods=['PATCH'])
```
mfcrocker: This should be DELETE not PATCH - we can deprecate or remove the old DELETE endpoint if this is the way we're going to do things now

mfcrocker (changes requested):
Would like the directory change and DELETE vs PATCH change done. The performance decision is one for you to make - happy to talk it through if you need to, but if you think the way its implemented is the best option keep it like that.

ras_party/batch_process/batch_request.py
creativeappdevamit: cool! moved `bach_resquest.py` to `/views`

ras_party/batch_process/batch_request.py
creativeappdevamit: as discussed the reason of batch was to avoid multiple patch.

ras_party/views/respondent_view.py
creativeappdevamit: Cool... changed to `DELETE` and removed the old delete endpoint

mfcrocker (approved):

PR #295, opened 2021-01-07
Chnage base logging level to debug

insacuri:
# Motivation and Context

# What has changed

# How to test?

# Links


jdfox (approved):

creativeappdevamit (commented):
Why are we changing the log level to debug?

insacuri:
The logging I added in the last PR isn't appearing in preprod.  Also, and more importantly, there are a number of other debug statements that have potentially useful information in them that are being suppressed.

It's better to have more logging then less, then delete any logging that is useless

PR #308, opened 2021-02-01
Auth deletion bug

creativeappdevamit:
# Motivation and Context
db session handler wraps the response which seems to create unhandled exception

# What has changed
forced user delete to return 500 and 404 status code.

# How to test?

# Links

### DIFF
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -88,7 +88,8 @@ def update_respondent_mark_for_deletion(email, session):
             session.query(Respondent).filter(Respondent.party_uuid == respondent.party_uuid) \
                 .update({Respondent.mark_for_deletion: True})
             return 'respondent successfully marked for deletion', 202
-        except (SQLAlchemyError, Exception):
+        except (SQLAlchemyError, Exception) as error:
+            logger.error('error with update respondent mark for deletion', error)
             return 'something went wrong', 500
     else:
         return 'respondent does not exist', 404
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -2,6 +2,8 @@ import logging
 import uuid
 
 import structlog
+from flask import jsonify
+from sqlalchemy.exc import SQLAlchemyError
 from werkzeug.exceptions import BadRequest, NotFound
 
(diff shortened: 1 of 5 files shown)


ras_party/controllers/respondent_controller.py
```diff
@@ -81,12 +83,15 @@ def update_respondent_mark_for_deletion(email, session):
...
-        raise NotFound("Respondent does not exist")
-    logger.info("Marking respondent for deletion", email=obfuscate_email(email))
-    session.query(Respondent).filter(Respondent.party_uuid == respondent.party_uuid) \
-        .update({Respondent.mark_for_deletion: True})
+    if respondent:
+        try:
+            session.query(Respondent).filter(Respondent.party_uuid == respondent.party_uuid) \
+                .update({Respondent.mark_for_deletion: True})
+            return 'respondent successfully marked for deletion', 202
+        except (SQLAlchemyError, Exception):
```
warrenbailey: It's probably worth logging the error here so we can investigate

test/test_respondent_controller.py
```diff
@@ -1494,11 +1494,11 @@ def test_put_change_respondent_account_status_bad_oauth2_response(self):
 
     def test_change_respondent_password_bad_auth_response(self):
         with patch('ras_party.controllers.account_controller.decode_email_token'), \
-             patch('ras_party.controllers.account_controller.current_app'), \
-             patch('ras_party.support.session_decorator.current_app.db'), \
-             patch('ras_party.controllers.account_controller.enrol_respondent_for_survey'), \
-             patch('ras_party.controllers.account_controller.NotifyGateway'), \
-             patch('ras_party.controllers.account_controller.OauthClient') as auth, \
+                patch('ras_party.controllers.account_controller.current_app'), \
```
warrenbailey, line 1497: did you mean to change this indentation?

warrenbailey (changes requested):

test/test_respondent_controller.py
creativeappdevamit: no but the auto format for pycharm did and it looks good now

ras_party/controllers/respondent_controller.py
creativeappdevamit: yes sure.. will update it.

warrenbailey (approved):

PR #311, opened 2021-02-22
Fix bug where 'verbose' removes data that is already there

insacuri:
# Motivation and Context
Currently, when you add `?verbose=true` to the end of `business/ref/<ref>` it actually removes a bunch of key details like the id, name, etc.  This PR fixes that and makes it so that verbose only adds more and takes nothing away!

# What has changed

# How to test?

# Links


insacuri:
Turns out I was mistaken.  The fields I thought were missing were actually there all along

PR #313, opened 2021-02-23
Remove search by party for ru_ref

insacuri:
# Motivation and Context
A search by ru_ref is always going to be a business, so it should exclusively be in the /businesses part of the API, otherwise we have 2 bits of code that do basically identical things.

This PR will temporarily map the old `/parties/type/B/ref/<ref>` to aid in the transition.  Once we're sure that all searches by ref work with this new combined way, we can work to make it into 1 single url.

The `/businesses/ref/<ref>` endpoint will lose the verbose flag.  Nothing used it, and we can always add it again later if we need to.  Simplifying and rationalising this API is important, so losing a minor feature isn't a huge deal.  The functionality of this endpoint will change slightly too.  Previously all the attributes were at the top level, mixed in with the name, trading_as, etc.  Now they will be in their own `attributes` section like it was in the previous `/parties...` url because it's much cleaner design.

# What has changed

# How to test?

# Links


test/test_party_controller.py
```diff
@@ -223,28 +205,43 @@ def test_put_business_sample_link_200(self):
...
         self._make_business_attributes_active(mock_business)
 
-        response = self.get_business_by_ref(mock_business['sampleUnitRef'], query_string={"verbose": "true"})
-
-        del mock_business['sampleSummaryId']
-        for x in mock_business:
-            self.assertIn(x, response)
+        ru_ref = mock_business['sampleUnitRef']
+        actual = self.get_business_by_ref(ru_ref)
+        expected = {'associations': [],
```
mfcrocker: Is this big enough to be included as a separate JSON file (as we've done elsewhere)?

mfcrocker (approved):

test/test_party_controller.py
insacuri: yeah that's a fair point.  At the time I was just happy to get to work that I didn't think about tidying it up ^^

PR #320, opened 2021-03-31
Share survey story 2

creativeappdevamit:
# What and why?
Share survey story 2 changes for party

# Trello
https://trello.com/c/n5Y8JhXP/559-s31-user-account-creation-share-surveys-background-functionality-to-enable-sharing-users-story-2


ras_party/controllers/queries.py
```diff
@@ -4,13 +4,40 @@
...
+    """
+    Query to return total pending share against businesses is and survey id
+    :param business_id: business party id
+    :param survey_id: survey id
+    :param session: db session
+    :return: the pending share
+    """
+    logger.info('Querying pending share by business_id and survey_id', business_id=business_id, survey_id=survey_id)
+    return session.query(PendingShares).filter(
+        PendingShares.business_id == business_id).filter(PendingShares.survey_id == survey_id)
```
insacuri, line 38: Chaining filters for a simple search seems odd to me (though correct me if I'm wrong about that).  Would something like this work?:
```python
return session.query(PendingShares).filter(
    and_(PendingShares.business_id == business_id,
             PendingShares.survey_id == survey_id))
```

ras_party/support/session_decorator.py
```diff
@@ -4,6 +4,7 @@
 import structlog
 from flask import current_app
 from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy import exc
```
insacuri: This still needed?

ras_party/models/models.py
```diff
@@ -360,3 +360,21 @@ class Enrolment(Base):
...
+    time_shared = Column(DateTime, default=datetime.datetime.utcnow)
+    Index('pending_shares_business_idx', business_id)
+    Index('pending_shares_email_address_idx', email_address)
+    Index('pending_shares_survey_idx', survey_id)
+    Index('pending_shares_time_shared_idx', time_shared)
+
+    __table_args__ = (
+        ForeignKeyConstraint(['business_id'],
+                             ['business.party_uuid']),
+        UniqueConstraint('email_address', 'business_id', 'survey_id', name='u_constraint'),
```
insacuri, line 383: More of a question than something that needs to be changed, but is this needed?  Looking at the docs you can define a composite primary key by doing mutiple `primary_key=True` like you've done here. or is there something I'm not understanding.  It's probably the latter as I've not done much DB stuff lately!

ras_party/views/share_survey_view.py
```diff
@@ -0,0 +1,84 @@
...
+    if username == config_username:
+        return config_password
+
+
+@share_survey_view.route('/share-survey-users-count', methods=['GET'])
+def share_survey_users():
+    """
+    Get total users count who are already enrolled and pending share survey against business id and survey id
+    """
+    payload = request.get_json()
```
insacuri: This feels wrong.  If I remember right, having json payloads in HTTP GET is bad practice.  Even though 'share survey' isn't an HTTP resource as such, it'd be better to have the url be `/share-survey-users-count?business_id=<id>&survey_id=<id>`

ras_party/views/share_survey_view.py
```diff
@@ -0,0 +1,84 @@
...
+    try:
+        business_id = payload['business_id']
+        survey_id = payload['survey_id']
+        # this is just to validate that the business_id exists
+        get_business_by_id(business_id)
```
insacuri, line 40: The responses for this endpoint aren't quite right either I think.  I'd expect a 400 if either wasn't supplied.  A 404 if the business/survey doesn't exist and a 200 on success.  It's more of a faff, but it also allows for more sensible errors to be given to the clients of this API

ras_party/views/share_survey_view.py
```diff
@@ -0,0 +1,84 @@
...
+            v = Validator(Exists(*expected))
+            if not v.validate(share):
+                logger.debug(v.errors)
+                return BadRequest(v.errors)
+        for pending_share in pending_shares:
+            share_survey_controller.pending_share_create(business_id=pending_share['business_id'],
+                                                         survey_id=pending_share['survey_id'],
+                                                         email_address=pending_share['email_address'])
+            # TODO: Add logic to send email
+        return make_response(jsonify({"created": "success"}), 201)
```
insacuri, line 85: Nice use of status codes ^^

ras_party/views/share_survey_view.py
```diff
@@ -0,0 +1,84 @@
...
+            "survey_id": "survey_id",
+            "email_address": "email_address"
+        }]
+     }
+    """
+    payload = request.get_json() or {}
+    try:
+        pending_shares = payload['pending_shares']
+        if len(pending_shares) == 0:
+            return BadRequest('Payload Invalid')
```
insacuri: Should be more specific here.  You can mention something like `pending_shares list is empty` or something

ras_party/views/share_survey_view.py
```diff
@@ -0,0 +1,84 @@
...
+                logger.debug(v.errors)
+                return BadRequest(v.errors)
+        for pending_share in pending_shares:
+            share_survey_controller.pending_share_create(business_id=pending_share['business_id'],
+                                                         survey_id=pending_share['survey_id'],
+                                                         email_address=pending_share['email_address'])
+            # TODO: Add logic to send email
+        return make_response(jsonify({"created": "success"}), 201)
+    except KeyError:
+        return BadRequest('Payload Invalid')
```
insacuri: Can be more specific here too.  This could be something like `pending_shares key not present in payload json`

insacuri (changes requested):
A few small changes but it's basically good to go ^^

ras_party/controllers/queries.py
creativeappdevamit: its basically the same thing first we filtering on business_id and then filtering on survey id. So when looking for survey id its a smaller subset

ras_party/support/session_decorator.py
creativeappdevamit: no.. will take it out

ras_party/models/models.py
creativeappdevamit: even I have not done it a lot so I am not sure. Although the sql script takes care of the unique constraint and I am not necessarily relying on sqlalchemy.

ras_party/views/share_survey_view.py
creativeappdevamit: yes you right ... this needs changing

ras_party/views/share_survey_view.py
creativeappdevamit: which eventually is an invalid payload.

ras_party/views/share_survey_view.py
creativeappdevamit: still an invalid payload.

ras_party/models/models.py
insacuri: Oki dokes, if it works then leave it as is.

ras_party/views/share_survey_view.py
insacuri: Yeah, but in this file it can be invalid for 2 different reasons.  Having 2 separate error messages instead of just 1 will make debugging easier down the line
EDIT:  The message could be: `Payload invalid: pending_shares list is empty` as another example

ras_party/controllers/queries.py
insacuri: Oki dokes, just leave it as is.  The comment was fairly minor ^^

ras_party/views/share_survey_view.py
creativeappdevamit: business/survey returns bad request. also include the request parameter 400 error. Will provide test cases to showcase the same.

ras_party/views/share_survey_view.py
creativeappdevamit: Ya sure changed it now

ras_party/views/share_survey_view.py
creativeappdevamit: made the changes

insacuri (approved):

insacuri (approved):

PR #322, opened 2021-05-07
Docker tag latest at the same time as the release

warrenbailey:
# What and why?
During the release step tag the generated docker image as both the release version and latest.

# How to test?

# Trello


insacuri (approved):

warrenbailey:
Status not reported but it has passed

PR #330, opened 2021-05-26
Share survey story 5

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/t8i7o3Pr/481-s35-user-account-creation-share-surveys-does-not-hold-account-story-5

### DIFF
--- a/ras_party/controllers/share_survey_controller.py
+++ b/ras_party/controllers/share_survey_controller.py
@@ -96,8 +96,7 @@ def get_unique_pending_shares(session):
     return [unique_batch_record.to_share_dict() for unique_batch_record in unique_batch_record]
 
 
-@with_db_session
-def validate_share_survey_token(token, session):
+def validate_share_survey_token(token):
     """
     Validates the share survey token and returns the pending shares against the batch number
     :param: token
@@ -114,7 +113,7 @@ def validate_share_survey_token(token, session):
     except (BadSignature, BadData):
         logger.exception("Bad token in validate_share_survey_token")
         raise NotFound("Unknown batch number in token")
-    return _get_share_survey_by_batch_number(batch_no, session)
+    return get_share_survey_by_batch_number(batch_no)
 
 
 @with_db_session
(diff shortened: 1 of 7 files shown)


ras_party/controllers/share_survey_controller.py
```diff
@@ -109,14 +114,22 @@ def validate_share_survey_token(token, session):
...
-    share_surveys = query_share_survey_by_batch_no(batch_no, session)
-    if len(share_surveys) == 0:
-        raise NotFound('Batch number does not exist')
-    return [share_survey.to_share_dict() for share_survey in share_surveys]
+    return _get_share_survey_by_batch_number(batch_no, session)
 
 
 @with_db_session
-def accept_share_survey(batch_no, session):
+def confirm_share_survey(batch_no, session):
```
insacuri, line 120: Why does this exist? Why can't you just call `accept_share_survey` and leave the annotation on that function

ras_party/views/share_survey_view.py
```diff
@@ -150,5 +150,15 @@ def confirm_pending_shares(batch_no):
...
+
+
+@share_survey_view.route('/share-survey/<batch_no>', methods=['GET'])
+def get_pending_share_with_batch_no(batch_no):
+    """
+    Confirms pending share survey
+    :param batch_no
+    """
+    response = get_share_survey_by_batch_number(batch_no)
+    return make_response(jsonify(response), 200)
```
insacuri, line 164: I'm pretty sure this can be replaced with 
```python
return jsonify(response)
```
as you're not doing anything fancy with the status

insacuri (changes requested):
A few small changes but mostly looks good

ras_party/controllers/share_survey_controller.py
creativeappdevamit: This one is for two different flow to reuse the code one initiates the session but as per the other scenario session is injected hence its has two. In doc string I have mentioned one initiates the db session one doesn't

ras_party/views/share_survey_view.py
creativeappdevamit: I am sure you are right. But with respect to other code keeping it consistent can we ignore it for now.

insacuri: yeah sure

ras_party/controllers/share_survey_controller.py
creativeappdevamit: 👍

insacuri (approved):

PR #335, opened 2021-06-08
Upgrade flask

mfcrocker:
# What and why?
Flask has had a major release, so this makes use of it. It also removes a couple of tests that we can't mock under Flask 2

# How to test?

# Trello
https://trello.com/c/W8y9rRfX/711-s36-upgrade-party-to-flask-v2xx


creativeappdevamit (approved):
Looks ok to me. I will try to test it on my namespace once the spinnaker is up.

PR #339, opened 2021-07-01
Making changes to make sure that multiple transfers works and becomes…

creativeappdevamit:
… shares with removing the originator

# What and why?

# How to test?

# Trello
https://trello.com/c/WseJBlY5/776-user-account-creation-handling-multiple-transfers-of-survey

### DIFF
--- a/ras_party/controllers/pending_survey_controller.py
+++ b/ras_party/controllers/pending_survey_controller.py
@@ -406,8 +406,8 @@ def send_pending_surveys_confirmation_email(pending_surveys_list, confirmation_e
         else:
             confirmation_email_template = 'share_survey_access_confirmation'
         business_list = []
-        business_id_list = [pending_surveys['business_id'] for pending_surveys in pending_surveys_list]
-        for business_id in set(business_id_list):
+        business_id_list = {pending_surveys['business_id'] for pending_surveys in pending_surveys_list}
+        for business_id in business_id_list:
             business = get_business_by_id(str(business_id))
             business_list.append(business['name'])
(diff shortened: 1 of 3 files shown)


ras_party/controllers/pending_survey_controller.py
```diff
@@ -404,8 +406,9 @@ def send_pending_surveys_confirmation_email(pending_surveys_list, confirmation_e
         else:
             confirmation_email_template = 'share_survey_access_confirmation'
         business_list = []
-        for survey in pending_surveys_list:
-            business = get_business_by_id(str(survey['business_id']))
+        business_id_list = [pending_surveys['business_id'] for pending_surveys in pending_surveys_list]
```
insacuri: If you change this to:
```python
business_id_list = {pending_surveys['business_id'] for pending_surveys in pending_surveys_list}
```
Doing a set comprehension will let you do the following on the next line.  It's minor but it's tidier.

```python
for business_id in business_id_list
```

ras_party/views/pending_survey_view.py
```diff
@@ -120,10 +120,13 @@ def post_pending_surveys():
     if not respondent:
         raise BadRequest('Originator unknown')
     batch_number = uuid.uuid4()
+    # logic to extract business list
+    business_id_list = [pending_surveys['business_id'] for pending_surveys in pending_surveys]
```
insacuri: same as my other comment, if you use set comprehension then you can drop the `set()` on the next line

insacuri (approved):
Minor change, but I won't block it

PR #344, opened 2021-07-12
Remove get party by ref code

insacuri:
# What and why?

# How to test?

# Trello


_infra/helm/party/Chart.yaml
```diff
@@ -18,4 +18,4 @@ version: 1.2.7
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 2.2.1
+appVersion: 2.2.2
```
warrenbailey: it might be worth going to 2.3.0 for this change

warrenbailey (approved):

PR #351, opened 2021-07-30
Added endpoint for pending surveys for response operation support

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/680HdmWq/778-s40-user-account-creation-sharing-and-transferring-rops-alterations

### DIFF
--- a/ras_party/controllers/pending_survey_controller.py
+++ b/ras_party/controllers/pending_survey_controller.py
@@ -23,6 +23,7 @@ from ras_party.controllers.queries import (
     query_business_respondent_by_respondent_id_and_business_id,
     query_enrolment_by_business_and_survey_and_status,
     query_pending_survey_by_batch_no,
+    query_pending_survey_by_shared_by,
     query_pending_surveys_by_business_and_survey,
     query_respondent_by_party_uuid,
 )
@@ -71,6 +72,28 @@ def get_users_enrolled_and_pending_survey_against_business_and_survey(business_i
     total_users = enrolled_users.count() + pending_survey_users.count()
     bound_logger.info(f"total users count {total_users}")
     return total_users
+
+
+@with_db_session
+def pending_survey_deletion(batch_no, session):
+    """
+    Delete pending survey record against batch number
+    :param batch_no: batch number
+    :type batch_no: uuid
+    :param session:
+    :type session:
+    """
+    logger.info("Retrieving pending share record against batch number", batch_no=str(batch_no))
+    pending_surveys = session.query(PendingSurveys).filter(PendingSurveys.batch_no == batch_no)
+    if pending_surveys.count() > 0:
+        try:
+            pending_surveys.delete()
+            return "pending surveys successfully deleted", 202
+        except (SQLAlchemyError, Exception) as error:
+            logger.error("error while deleting pending surveys", error)
+            return "something went wrong", 500
+    else:
+        return "pending surveys does not exists.", 404
 
 
 @with_db_session
@@ -330,6 +353,21 @@ def get_pending_survey_by_batch_number(batch_number, session):
     pending_surveys = query_pending_survey_by_batch_no(batch_number, session)
     if len(pending_surveys) == 0:
         raise NotFound("Batch number does not exist")
+    return [pending_surveys.to_pending_surveys_dict() for pending_surveys in pending_surveys]
+
+
+@with_db_session
+def get_pending_survey_by_originator_respondent_id(respondent_party_id, session):
+    """
+    gets list of pending surveys against the respondent party id
+    :param respondent_party_id: respondent party id
+    :type batch_number: uuid
+    :param session: db session
+    :type session: db session
+    :return: list of pending share surveys
+    :rtype: list
+    """
+    pending_surveys = query_pending_survey_by_shared_by(respondent_party_id, session)
     return [pending_surveys.to_pending_surveys_dict() for pending_surveys in pending_surveys]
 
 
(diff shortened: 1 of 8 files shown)


ras_party/views/pending_survey_view.py
```diff
@@ -200,8 +202,87 @@ def confirm_pending_shares(batch_no):
...
+    payload = request.get_json() or {}
+    if "batch_number" not in payload:
+        logger.error("Invalid request")
+        raise BadRequest("Invalid request - batch_number missing")
+    batch_number = payload["batch_number"]
+    logger.info("retrieving pending surveys against batch number", batch_number=batch_number)
+    response = get_pending_survey_by_batch_number(batch_number)
+    logger.info("retrieving originator email address for resend email", batch_number=batch_number)
+    originator = get_respondent_by_id(str(response[0]["shared_by"]))
+    if not originator:
```
insacuri: This function either returns the dict (if it exists) or raises either a BadRequest or NotFound exception (depending on what is wrong).  That means the code in this if statement will never be hit.  I'd either remove it if you're okay with the raised exceptions from the function, or catch and handle any exceptions yourself.

If you do the former, then you might need to change your code to handle a 400 or 404 coming back from this endpoint

ras_party/views/pending_survey_view.py
```diff
@@ -200,8 +202,87 @@ def confirm_pending_shares(batch_no):
...
+    if "is_transfer" in response[0]:
+        verification_url = PublicWebsite().transfer_survey(batch_number)
+        existing_user_email_template = "transfer_survey_access_existing_account"
+        new_user_email_template = "transfer_survey_access_new_account"
+    else:
+        verification_url = PublicWebsite().share_survey(batch_number)
+        existing_user_email_template = "share_survey_access_existing_account"
+        new_user_email_template = "share_survey_access_new_account"
+    try:
+        get_respondent_by_email(respondent_email_address)
```
insacuri, line 269: Might be worth putting a comment saying something like 'Ignoring the return value as we're only testing if they exist' or something to that effect as it looks weird to use this function and discard the return value at first glance

ras_party/views/pending_survey_view.py
```diff
@@ -200,8 +202,87 @@ def confirm_pending_shares(batch_no):
...
+    :param batch_no: batch number
+    :type batch_no: uuid
+    """
+    logger.info("Attempting to delete pending surveys record against batch number", batch_no=batch_no)
+    response = pending_survey_deletion(batch_no)
+    return response
+
+
+@pending_survey_view.route("/pending-surveys/resend-email", methods=["POST"])
+def resend_pending_surveys_email():
```
insacuri, line 240: Very minor request so ignore if you want, but can you put some of this code into separate functions?  This function feels pretty long.

ras_party/views/pending_survey_view.py
```diff
@@ -200,8 +202,87 @@ def confirm_pending_shares(batch_no):
...
+    }
+    """
+    logger.info("Attempting to resend email for pending survey")
+    payload = request.get_json() or {}
+    if "batch_number" not in payload:
+        logger.error("Invalid request")
+        raise BadRequest("Invalid request - batch_number missing")
+    batch_number = payload["batch_number"]
+    logger.info("retrieving pending surveys against batch number", batch_number=batch_number)
+    response = get_pending_survey_by_batch_number(batch_number)
```
insacuri: `response` isn't a great name for this I feel. maybe `pending_surveys` ? or something like that

ras_party/controllers/pending_survey_controller.py
```diff
@@ -73,6 +74,28 @@ def get_users_enrolled_and_pending_survey_against_business_and_survey(business_i
     return total_users
 
 
+@with_db_session
+def pending_survey_deletion(batch_no, session):
+    """
+    Delete pending survey record against batch number
+    :param batch_no: batch number
+    :type batch_no: uuid
```
insacuri: Is the batch no a UUID object or a string?

ras_party/controllers/pending_survey_controller.py
```diff
@@ -333,6 +356,21 @@ def get_pending_survey_by_batch_number(batch_number, session):
     return [pending_surveys.to_pending_surveys_dict() for pending_surveys in pending_surveys]
 
 
+@with_db_session
+def get_pending_survey_by_originator_respondent_id(respondent_party_id, session):
+    """
+    gets list of pending surveys against the respondent party id
+    :param respondent_party_id: respondent party id
+    :type batch_number: uuid
```
insacuri: There is no batch number param.  Also is it a UUID object or a string?  Either way, can you add it as a type hint instead of a docstring

insacuri (changes requested):
Looks good, just a few small changes and it should be good to go

ras_party/views/pending_survey_view.py
creativeappdevamit: right. I missed the exception in the calling method. Will remove if statement.

ras_party/views/pending_survey_view.py
creativeappdevamit: I will leave it for now, it looks ok to me. If thats ok.

insacuri: yeah no problem

ras_party/controllers/pending_survey_controller.py
creativeappdevamit: str will change

ras_party/controllers/pending_survey_controller.py
creativeappdevamit: str will change

ras_party/views/pending_survey_view.py
creativeappdevamit: done

insacuri (approved):

PR #355, opened 2021-09-10
Adding call to send account deletion confirmation email

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/h8nV6an7/883-user-account-creation-account-deletion-request

### DIFF
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -2,6 +2,7 @@ import logging
 import uuid
 
 import structlog
+from flask import current_app
 from sqlalchemy.exc import SQLAlchemyError
 from werkzeug.exceptions import BadRequest, NotFound
 
@@ -9,6 +10,7 @@ from ras_party.controllers.account_controller import (
     change_respondent,
     get_single_respondent_by_email,
 )
+from ras_party.controllers.notify_gateway import NotifyGateway
 from ras_party.controllers.queries import (
     query_respondent_by_email,
     query_respondent_by_names_and_emails,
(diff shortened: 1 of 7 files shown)


ras_party/controllers/respondent_controller.py
```diff
@@ -123,6 +125,31 @@ def delete_respondents_marked_for_deletion(session):
         session.query(BusinessRespondent).filter(BusinessRespondent.respondent_id == respondent.id).delete()
         session.query(PendingEnrolment).filter(PendingEnrolment.respondent_id == respondent.id).delete()
         session.query(Respondent).filter(Respondent.id == respondent.id).delete()
+        send_account_deletion_confirmation_email(respondent.email_address, respondent.first_name)
+
+
+def send_account_deletion_confirmation_email(email_address, name):
```
insacuri: Put type hints here
```python
def send_account_deletion_confirmation_email(email_address: str, name: str):
```

ras_party/controllers/respondent_controller.py
```diff
@@ -123,6 +125,31 @@ def delete_respondents_marked_for_deletion(session):
...
+    """
+    Sends email notification for account deletion confirmation.
+    :param email_address:
+    :type email_address:
```
insacuri: Can you tidy this bit up?  The 2 types aren't needed (as they're typehints in the function) and rtype can go too as this function doesn't return anything

insacuri (changes requested):
2 small changes, but looks good

ras_party/controllers/respondent_controller.py
creativeappdevamit: done

ras_party/controllers/respondent_controller.py
creativeappdevamit: done

insacuri (approved):

PR #359, opened 2021-12-03
updates to improve search results

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/XjLiwqc8/1004-spike-response-operations-performance-reporting-unit-page-1-2-days

### DIFF
--- a/ras_party/views/business_view.py
+++ b/ras_party/views/business_view.py
@@ -88,7 +88,8 @@ def get_party_by_search():
     page = int(request.args.get("page", default=1))
     limit = int(request.args.get("limit", default=100))
     is_ru_ref_search = eval(request.args.get("ru_search", default="False"))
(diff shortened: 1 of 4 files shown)


ras_party/views/business_view.py
```diff
@@ -88,7 +88,8 @@ def get_party_by_search():
     page = int(request.args.get("page", default=1))
     limit = int(request.args.get("limit", default=100))
     is_ru_ref_search = eval(request.args.get("ru_search", default="False"))
+    max_rec = int(request.args.get("max_rec", default=10000))
```
mybigopenbox, line 91: why 10,000. seems high as 9,999 returned records would be a lot of pges

mybigopenbox (approved):

PR #362, opened 2021-12-10
making response ops for account email change consistent

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/yk676X1g/1225-change-rops-resend-email-to-use-the-new-template

### DIFF
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -58,6 +58,7 @@ logger = structlog.wrap_logger(logging.getLogger(__name__))
 
 NO_RESPONDENT_FOR_PARTY_ID = "There is no respondent with that party ID"
 EMAIL_VERIFICATION_SENT = "A new verification email has been sent"
+ACCOUNT_EMAIL_CHANGE_VERIFICATION_SENT = "A new account email change verification email has been sent"
 
 
 # flake8: noqa: C901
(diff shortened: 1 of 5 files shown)


ras_party/controllers/account_controller.py
```diff
@@ -58,6 +58,7 @@
 
 NO_RESPONDENT_FOR_PARTY_ID = "There is no respondent with that party ID"
 EMAIL_VERIFICATION_SENT = "A new verification email has been sent"
+ACCOUNT_EMAIL_CHANGE_VERIFICATION_SENT = "A new account email change verification email has been sent"
```
insacuri: This reads kinda weird.... maybe
`A verification email has been sent to change the email address on the account` or something like that?  My suggestion kinda sucks so don't use that!  Can you ask for possibly better wording for it from someone else?

ras_party/views/account_view.py
```diff
@@ -76,6 +76,12 @@ def resend_verification_email(party_uuid):
     return make_response(jsonify(response), 200)
 
 
+@account_view.route("/resend-account-email-change-notification/<party_uuid>", methods=["POST"])
+def resend_account_email_change_verification(party_uuid):
+    response = account_controller.resend_account_email_change_verification_email_by_uuid(party_uuid)
+    return make_response(jsonify(response), 200)
```
insacuri, line 82: This can just be `jsonify(response)`

insacuri (changes requested):
2 small changes but overall, looks good

ras_party/views/account_view.py
creativeappdevamit: for consistency we should add the http code I believe

ras_party/controllers/account_controller.py
creativeappdevamit: 👍

insacuri (approved):

PR #363, opened 2022-03-02
Re-enable respondent

arroyoAle:
# What and why?
Adding logic to enable respondents to be re-enabled in case when they are re-enabled in rOps
# How to test?

# Trello
https://trello.com/c/8mXH30EL/1396-bug-enrolment-status-change-does-not-get-propagated-to-case-service


creativeappdevamit (changes requested):
can we provide some test against the changes please

ras_party/controllers/account_controller.py
```diff
@@ -975,7 +988,8 @@ def request_survey(survey_id):
 def request_casegroups_for_business(business_id):
     logger.info("Retrieving casegroups for business", business_id=business_id)
     url = f'{current_app.config["CASE_URL"]}/casegroups/partyid/{business_id}'
-    response = Requests.get(url)
+    auth = (current_app.config["SECURITY_USER_NAME"], current_app.config["SECURITY_USER_PASSWORD"])
```
creativeappdevamit, line 991: why we changed this?

Pipfile
```diff
@@ -32,3 +32,4 @@ flake8 = "*"
 flask-testing = "*"
 isort = "*"
 pytest-cov = "*"
+responses = "*"
```
creativeappdevamit, line 35: why did we add this?

creativeappdevamit (changes requested):

Pipfile
insacuri: For the mocks in the unit tests

ras_party/controllers/account_controller.py
insacuri: It's a long story... We used responses to sort the mocks in our new tests.  It worked in pycharm but for some reason not when you ran it via command line.  We managed to get it working by removing the custom (and silly I think) wrapper around requests and just use it like we do everywhere else.
That then broke a bunch of other tests because I think things were working by luck rather then because proper mocking had been done, after fixing those tests we finally got everything passing!

Basically, it's identical to how it was before and more closely matches how we do http calls in all our other python repos.

creativeappdevamit (approved):

ras_party/controllers/account_controller.py
arroyoAle: The previous implementation was using a requests class made by us with less functionality that was also making the tests fail

Pipfile
arroyoAle: to mock the requests in the unit tests

PR #364, opened 2022-03-10
Setup party to check verification tokens used in email are valid

arroyoAle:
# What and why?
Setup party to check verification tokens used in email are valid. For now it is for the reset password email and can be used for others in future
# How to test?

# Trello


ras_party/models/models.py
```diff
@@ -283,6 +283,7 @@ class Respondent(Base):
     telephone = Column(Text)
     mark_for_deletion = Column(Boolean, default=False)
     created_on = Column(DateTime, default=datetime.datetime.utcnow)
+    verification_tokens = Column(ARRAY(Text))
```
insacuri: You can have multiple tokens live at once?  That seems kinda weird to me.

EDIT:  If the respondent needs a new token then you can create a new one and invalidate the old one (by overwriting it)

mfcrocker (changes requested):
I don't believe this has been implemented in a way that adheres to RESTful principles. A `PUT` should replace the resource on the server with the one provided in the body - the endpoint that's been created here is a mashup between a `POST` and a `DELETE` depending on whether the verification token is already stored.

The endpoint is also not really named in a way that references resources (which is probably an indication that something isn't properly RESTful).

Thankfully, this shouldn't be too difficult to fix and just needs us to split things in a slightly different way. I'd suggest 2 endpoints (on top of having verification tokens returned in the GET respondent by email endpoint):

`POST /respondents/{respondent_uuid}/verification_tokens`

`{
   "token": "<token>"
}`

This would add a verification token to a user

`DELETE /respondents/{respondent_uuid}/verification_tokens/{token}`

This would remove a verification token from a user

All in all this is good work and you can almost certainly keep most of the code here, but I think we should be striving to keep our APIs clean and compliant with REST principles where possible (and rethinking what we're doing if we can't do that).

ras_party/models/models.py
insacuri: Also this column name doesn't seem all that descriptive for me.  The token we're putting in this column today is for password reset only right?  The current name implies you can put tokens for whatever verification purpose in here which doesn't feel right

ras_party/controllers/account_controller.py
```diff
@@ -401,6 +403,50 @@ def verify_token(token, session):
...
+    respondent = query_respondent_by_party_uuid(respondent_id, session)
+    if not respondent:
+        logger.info("Respondent with party id does not exist", respondent_id=respondent_id)
+        raise NotFound("Respondent id does not exist")
+    if not respondent.password_verification_token:
+        logger.info("Verification token not found")
+        raise NotFound("Verification token not found")
+    if not token:
+        logger.info("Verification token not received")
+        raise NotFound("Verification token not received")
```
mfcrocker: This would be a 400 Bad Request

```suggestion
        raise BadRequest("Verification token not received")
```

mfcrocker (approved):
Looks good, 1 tiny thing

PR #365, opened 2022-03-30
Add delete by sample summary id endpoint

insacuri:
# What and why?

# How to test?

# Trello


insacuri:
Just tried it out and the performance is bad, but passable, at 11 seconds for 40k records.  The equivalent party endpoint does it in 0.5 seconds so I'm going to experiment a little to see if we can get it closer to that.

mfcrocker (approved):

PR #367, opened 2022-04-06
Apply restrictions to forgotten password journey.

arroyoAle:
# What and why?
Set up party to allow restrictions on amount of password reset attempts
# How to test?

# Trello

### DIFF
--- a/API.md
+++ b/API.md
@@ -1094,7 +1094,7 @@ Returns:
 ### Update Respondent Info
 
 * `PUT /party-api/v1/respondents/id/<respondent_id>`
-  * Updates a respondents details (first name, last name, telephone and email) based on `party_uuid` from the respondents table.
+  * Updates a respondent's details (first name, last name, telephone and email) based on `party_uuid` from the respondents table.
 
 #### Example JSON payload
 
@@ -1109,9 +1109,21 @@ Returns:
 ```
 
 ---
-### Add Respondent Tokens
-* `POST /party-api/v1/respondents/<respondent_id>/verification-tokens`
-  * Updates a respondents valid verification tokens based on `party_uuid` deduced from the `email` passed.
+### Get Respondent Token
+* `GET /party-api/v1/respondents/<respondent_id>/verification-token`
+  * Retrieves a respondent's valid verification token based on `party_uuid`
+
+### Success
+```json
+{
+  "token": "<token>"
+}
+```
+---
+  
+### Add Respondent Token
+* `POST /party-api/v1/respondents/<respondent_id>/verification-token`
+  * Updates a respondent's valid verification token based on `party_uuid`
 
 #### Example JSON payload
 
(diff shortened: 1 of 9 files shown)


creativeappdevamit (approved):

API.md
```diff
@@ -1122,11 +1134,43 @@ Returns:
...
 
 ---
+### Get Password Reset Counter
+* `GET /party-api/v1/respondents/<respondent_id>/password-reset-counter`
+  * Retrieves a respondent's password reset counter based on the `party_uuid`
+
+### Success
+```json
+{
+  "counter": <counter>
```
insacuri: Needs quotes around the <counter> for it to be valid json

API.md
```diff
@@ -1122,11 +1134,43 @@ Returns:
...
+}
+```
+---
+
+---
+### Put Password Reset Counter
+* `PUT /party-api/v1/respondents/<respondent_id>/password-reset-counter`
```
insacuri: Why do you need this?  If a user attempts to get a password reset email then it should just increment itself without something external telling it to increment.  Or am I missing something?

ras_party/views/account_view.py
```diff
@@ -164,14 +164,41 @@ def post_pending_survey_respondent():
...
+@account_view.route("/respondents/<respondent_id>/password-verification-token/<token>", methods=["DELETE"])
+def delete_password_verification_token(respondent_id, token):
     account_controller.delete_respondent_password_token(respondent_id, token)
     return make_response(jsonify({"message": "Successfully removed token"}), 200)
+
+
+@account_view.route("/respondents/<respondent_id>/password-reset-counter", methods=["GET"])
+def get_password_reset_counter(respondent_id):
+    counter = account_controller.get_password_counter(respondent_id)
+    return make_response(jsonify({"counter": counter}), 200)
```
insacuri: for all of these simple endpoints that have 200 return values , doing:
```python
return jsonify(<insert dict here>)
```
will suffice.

I think (but you'll have to check... maybe in this repo or collection instrument?) you can do the same for a non-200 value, where it looks something like:
```python
return jsonify(<insert dict here>), 400
```

insacuri (changes requested):
A few small comments and one design question, but overall it looks good

ras_party/views/account_view.py
```diff
@@ -164,14 +164,41 @@ def post_pending_survey_respondent():
...
 
-@account_view.route("/respondents/<respondent_id>/password-verification-tokens", methods=["POST"])
-def post_password_verification_tokens(respondent_id):
+@account_view.route("/respondents/<respondent_id>/password-verification-token", methods=["GET"])
+def get_password_verification_token(respondent_id):
+    token = account_controller.get_respondent_password_token(respondent_id)
+    return make_response(jsonify({"token": token}), 200)
+
+
+@account_view.route("/respondents/<respondent_id>/password-verification-token", methods=["POST"])
```
insacuri, line 175: For the future, doing it this way is naughty, but I don't think it's a huge deal for this bit of work.
If you change the name of url in 2 services without allowing for backwards compatability (in this case, allowing for `-tokens` AND `-token`) then there will be a non-0 amount of time where the bit of functionality won't work.

For this one it's acceptable becuase it's for something that isn't constantly in use and will only stop working for a second.  But be more considerate in future as this'll bite you if it's a more core bit of functionality.

insacuri: For this one and the other one, can you add the old version as a route above it:

```python
@account_view.route("/respondents/<respondent_id>/password-verification-token", methods=["POST"])
@account_view.route("/respondents/<respondent_id>/password-verification-tokens", methods=["POST"])
def ....():
```
Put a comment somewhere around it that the `-tokens` route is deprecated
Also update the API documentation with the old route and also mention that it's deprecated.

insacuri (approved):

PR #368, opened 2022-05-04
Remove tokens endpoints

arroyoAle:
# What and why?
Removes two deprecated verification token endpoints which have been replaced with the correct endpoints
# How to test?

# Trello

### DIFF
--- a/API.md
+++ b/API.md
@@ -1122,8 +1122,6 @@ Returns:
 ---
   
 ### Add Respondent Token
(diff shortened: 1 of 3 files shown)


LJBabbage (commented):
Can you confirm where this would have been called from? I reckon it would have only been frontstage which looks fine

LJBabbage (approved):
Confirmed it is just frontstage

PR #377, opened 2022-10-20
BUG - rOps sampled businesses display 500 error in RU tab

SteveScorfield:
# What and why?
Two types of ru ref error are present and both return a 404. These are when a ru ref isn't present in the database and the ru ref is present but does not have a collection exercise. The latter is apart of the bug that needs to be addressed. The 404 for this error has been changed to a 400 so as to distinguish between the two when returning the error message with in rOps.

# How to test?

1. Create a new CE
2. Load a sample file with a company RU ref that does not exist yet for a survey
3. Don't set survey to live
4. Go into Reporting Units tab
5. Load RU of sampled, but not set ready for live business
6. Click on RU to load business details
7. A specific error page should be present for the failed RU ref
8. Then in the above url, change the ru ref to one that doesn't exist
9. A different error page should then show

# Trello
https://trello.com/c/wB7Vtrxr/2071-bug-rops-sampled-businesses-display-500-error-in-ru-tab-track-time


SteveScorfield:
[BUG - rOps sampled businesses display 500 error in RU tab (track time)](https://trello.com/c/wB7Vtrxr/2071-bug-rops-sampled-businesses-display-500-error-in-ru-tab-track-time)

insacuri (approved):

PR #385, opened 2022-12-23
Update container to 3.10

LJBabbage:
# What and why?
On reverting party back to Python 3.10 the Dockerfile was missed
# How to test?
Run all the apps with this version of PR of Party and make sure it works
# Trello


SteveScorfield (approved):
Looks good to me

PR #389, opened 2023-04-18
RAS-717 Update auto scalling to v2

LJBabbage:
# What and why?
Update autoscaling from v2beta2 to v2
# How to test?
Deploy via helm. I initial built a whole release then deleted the services associated in the charts. You will need to update your [values.yaml](https://github.com/ONSdigital/ras-party/blob/main/_infra/helm/party/values.yaml) file making sure env, namespace, project, topic and autoscaling is true

```
kubectl delete deployment party --namespace <your namespace>
```

This should then let you deploy with 

```
helm install party ./_infra/helm/party
```

Give it a couple of minutes as HPA takes a while, then click on party on gcp and you should see HPA autoscale with a green tick 

image

# Trello

### DIFF
--- a/_infra/helm/party/Chart.yaml
+++ b/_infra/helm/party/Chart.yaml
@@ -14,9 +14,9 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.4.24
+version: 2.4.25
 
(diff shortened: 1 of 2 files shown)


matthew-robinson-ons (approved):
Deployed via helm and version V2 of HPA is working.

PR #390, opened 2023-04-19
RAS-720 Update Cron from v1beta to v1

LJBabbage:
# What and why?
This PR updates CronJob from v1beta to v1 as the former is depreciated.

# How to test?
Deploy via helm. I initial built a whole release then deleted the services associated in the charts. You will need to update your [values.yaml](https://github.com/ONSdigital/ras-party/blob/main/_infra/helm/party/values.yaml) file making sure env, namespace, project, topic and autoscaling is true

```
kubectl delete deployment party --namespace <your namespace>
```

This should then let you deploy with 

```
helm install party ./_infra/helm/party
```

Now go to GCP workloads and make sure the Cronjobs party-scheduler-delete-respondents and party-scheduler-remove-expired-pending-surveys are created in your namespace with v1. On your terminal you should also not getting any warnings (you did before this change)

# Trello

### DIFF
--- a/_infra/helm/party/Chart.yaml
+++ b/_infra/helm/party/Chart.yaml
@@ -14,9 +14,9 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.4.24
+version: 2.4.25
 
 # This is the version number of the application being deployed. This version number should be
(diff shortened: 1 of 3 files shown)


matthew-robinson-ons (approved):
Deployed via helm and both cron jobs appear in GCP correctly under V1

PR #392, opened 2023-05-05
uppgrde python version, rebuild lock file, set gunicorn workers to 8

JacobHarrisonONS:
# What and why?
Upgrading back to python 3.11 and setting gunicorn back to 8 workers

# How to test?
deploy and sure everything still works correctly

# Jira
https://jira.ons.gov.uk/browse/RAS-749


matthew-robinson-ons (approved):
All looks good no errors seen in dev and party running as expected. There was a comment on the card around testing performance under large sample load but that will need to be checked in pre-prod.

PR #393, opened 2023-05-11
revert python upgrade

JacobHarrisonONS:
# What and why?
revert python upgrade

# How to test?

# Trello


matthew-robinson-ons (approved):
Checked changes and PR reverts all previous commits

PR #397, opened 2023-07-13
Temporarily suspend delete cronjob and bump requests from 2.30.0 to 2.31.0

pricem14pc:
# What and why?

- temporarily suspend the deletion cronjobs during deployment
- bump requests from 2.30.0 to 2.31.0

# How to test

- deploy to an env and ensure the `party-scheduler-delete-respondents` cronjob is suspended when deployed (using `/deploy` in this PR may not deploy the PR's tgz of this updated Helm chart change, one alternative is to manually deploy it using Helm from a local branch)
- run the acceptance tests to ensure the bump requests to 2.31.0 is ok

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,11 +18,11 @@
     "default": {
         "alembic": {
             "hashes": [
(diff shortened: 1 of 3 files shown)


JacobHarrisonONS (approved):
Cron is suspended and acceptance tests pass.

PR #400, opened 2023-08-30
Adding password reset email change fix

arroyoAle:
# What and why?
This fixes the case where a user changes email while having requested a forgot password link by deleting the password reset token in the database when the party email is changed

# How to test?
- deploy to environment
- follow the `forgot password` path for an account
- check the token is in the database
- change the user email and verify the change through the link provided 
- check the token has been removed from the database and the counter is reset
- 
# Jira
https://jira.ons.gov.uk/browse/RAS-840

### DIFF
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -723,9 +723,18 @@ def put_email_verification(token, tran, session):
         logger.info("Attempting to find respondent by pending email address")
         # When changing contact details, unverified new email is in pending_email_address
         respondent = query_respondent_by_pending_email(email_address, session)
+        print("here1")
 
         if respondent:
             update_verified_email_address(respondent, tran)
+            print("here")
+            if not respondent.pending_email_address and respondent.password_verification_token:
+                # Reset password token fields in the database since they are linked to the old email
+                print("here1")
+                delete_respondent_password_verification_token(respondent.party_uuid, session)
+                print("here2")
+                reset_password_reset_counter(respondent.party_uuid, session)
+                print("here3")
         else:
             logger.info("Unable to find respondent by pending email")
             raise NotFound("Unable to find user while checking email verification token")
@@ -745,15 +754,6 @@ def put_email_verification(token, tran, session):
 
         # We set the user as verified on the OAuth2 server.
         set_user_verified(email_address)
-
-    if not respondent.pending_email_address and respondent.password_verification_token:
-        if not respondent.password_verification_token:
-            logger.info("Verification token not found")
-            raise NotFound("Verification token not found")
-
-        # Reset password token fields in the database since they are linked to the old email
-        delete_respondent_password_verification_token(respondent.party_uuid, session)
-        reset_password_reset_counter(respondent.party_uuid, session)
 
     return respondent.to_respondent_dict()
 
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -723,18 +723,13 @@ def put_email_verification(token, tran, session):
         logger.info("Attempting to find respondent by pending email address")
         # When changing contact details, unverified new email is in pending_email_address
         respondent = query_respondent_by_pending_email(email_address, session)
-        print("here1")
 
         if respondent:
             update_verified_email_address(respondent, tran)
-            print("here")
             if not respondent.pending_email_address and respondent.password_verification_token:
                 # Reset password token fields in the database since they are linked to the old email
-                print("here1")
                 delete_respondent_password_verification_token(respondent.party_uuid, session)
-                print("here2")
                 reset_password_reset_counter(respondent.party_uuid, session)
-                print("here3")
         else:
             logger.info("Unable to find respondent by pending email")
             raise NotFound("Unable to find user while checking email verification token")
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -746,6 +746,10 @@ def put_email_verification(token, tran, session):
         # We set the user as verified on the OAuth2 server.
         set_user_verified(email_address)
 
+    # Delete password verification token and reset the counter
+    delete_respondent_password_token(respondent.party_uuid, respondent.password_verification_token)
+    reset_password_counter(respondent.party_uuid)
+
     return respondent.to_respondent_dict()
 
 
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -746,9 +746,14 @@ def put_email_verification(token, tran, session):
         # We set the user as verified on the OAuth2 server.
         set_user_verified(email_address)
 
-    # Delete password verification token and reset the counter
-    delete_respondent_password_token(respondent.party_uuid, respondent.password_verification_token)
-    reset_password_counter(respondent.party_uuid)
+    if not respondent.pending_email_address and respondent.password_verification_token:
+        if not respondent.password_verification_token:
+            logger.info("Verification token not found")
+            raise NotFound("Verification token not found")
+
+        # Reset password token fields in the database since they are linked to the old email
+        delete_respondent_password_verification_token(respondent.party_uuid, session)
+        reset_password_reset_counter(respondent.party_uuid, session)
 
     return respondent.to_respondent_dict()
 
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -726,7 +726,7 @@ def put_email_verification(token, tran, session):
 
         if respondent:
             update_verified_email_address(respondent, tran)
-            if not respondent.pending_email_address and respondent.password_verification_token:
+            if respondent.password_verification_token:
                 # Reset password token fields in the database since they are linked to the old email
                 delete_respondent_password_verification_token(respondent.party_uuid, session)
                 reset_password_reset_counter(respondent.party_uuid, session)
--- a/test/party_client.py
+++ b/test/party_client.py
@@ -435,30 +435,14 @@ class PartyTestClient(TestCase):
         response = self.client.delete(f"/party-api/v1/pending-surveys/{batch_no}", headers=self.auth_headers)
         self.assertStatus(response, expected_status)
 
-    def get_password_verification_token(self, party_id, expected_status=200):
-        response = self.client.get(f"/party-api/v1/respondents/{party_id}/password-verification-token")
-        self.assertStatus(response, expected_status)
-        return response.json
-
-    def post_password_verification_token(self, party_id, payload, expected_status=200):
-        response = self.client.post(
-            f"/party-api/v1/respondents/{party_id}/password-verification-token", json=payload, headers=self.auth_headers
-        )
-        self.assertStatus(response, expected_status, response.json)
-
-    def delete_password_verification_token(self, party_id, token):
+    def delete_password_verification_token(self, party_id, token, expected_status=200):
         response = self.client.delete(
             f"/party-api/v1/respondents/{party_id}/password-verification-token/{token}", headers=self.auth_headers
         )
-        self.assertStatus(response, 200)
+        self.assertStatus(response, expected_status)
 
-    def get_password_reset_counter(self, party_id):
-        response = self.client.get(f"/party-api/v1/respondents/{party_id}/password-reset-counter")
-        self.assertStatus(response, 200)
-        return response.json
-
-    def reset_password_reset_counter(self, party_id):
+    def reset_password_reset_counter(self, party_id, expected_status=200):
         response = self.client.delete(
             f"/party-api/v1/respondents/{party_id}/password-reset-counter", headers=self.auth_headers
         )
-        self.assertStatus(response, 200)
+        self.assertStatus(response, expected_status)
--- a/test/party_client.py
+++ b/test/party_client.py
@@ -434,3 +434,31 @@ class PartyTestClient(TestCase):
     def delete_pending_surveys_with_batch_no(self, batch_no, expected_status=202):
         response = self.client.delete(f"/party-api/v1/pending-surveys/{batch_no}", headers=self.auth_headers)
         self.assertStatus(response, expected_status)
+
+    def get_password_verification_token(self, party_id, expected_status=200):
+        response = self.client.get(f"/party-api/v1/respondents/{party_id}/password-verification-token")
+        self.assertStatus(response, expected_status)
+        return response.json
+
+    def post_password_verification_token(self, party_id, payload, expected_status=200):
+        response = self.client.post(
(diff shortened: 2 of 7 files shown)


ras_party/controllers/account_controller.py
```diff
@@ -746,6 +746,15 @@ def put_email_verification(token, tran, session):
         # We set the user as verified on the OAuth2 server.
         set_user_verified(email_address)
 
+    if not respondent.pending_email_address and respondent.password_verification_token:
+        if not respondent.password_verification_token:
```
LJBabbage, line 750: Is this reachable code? As the above means respondent.password_verification_token has to be True

ras_party/controllers/account_controller.py
```diff
@@ -746,6 +746,15 @@ def put_email_verification(token, tran, session):
         # We set the user as verified on the OAuth2 server.
         set_user_verified(email_address)
 
+    if not respondent.pending_email_address and respondent.password_verification_token:
```
LJBabbage: This feels in the wrong place and order. The issue materializes because the email in the token has the original email and due to the pending becoming the new email makes it makes it invalid. With that in mind it should be there and not here. Having it here puts in on the original verification path which isn't valid. 

It should be done inside update_verified_email_address or alternatively after line 727

ras_party/controllers/account_controller.py
LJBabbage: Is this reachable code? As the above means respondent.password_verification_token has to be True

ras_party/controllers/account_controller.py
```diff
@@ -746,6 +746,15 @@ def put_email_verification(token, tran, session):
...
+        if not respondent.password_verification_token:
+            logger.info("Verification token not found")
+            raise NotFound("Verification token not found")
+
+        # Reset password token fields in the database since they are linked to the old email
+        delete_respondent_password_verification_token(respondent.party_uuid, session)
```
LJBabbage: I do wonder if these 2 are intrinsically linked and should have a parent function, maybe worth checking other usages in the code

test/party_client.py
```diff
@@ -434,3 +434,31 @@ def get_pending_surveys_with_originator_party_id(self, party_id, expected_status
     def delete_pending_surveys_with_batch_no(self, batch_no, expected_status=202):
         response = self.client.delete(f"/party-api/v1/pending-surveys/{batch_no}", headers=self.auth_headers)
         self.assertStatus(response, expected_status)
+
+    def get_password_verification_token(self, party_id, expected_status=200):
```
LJBabbage: Is this used? I can't see where

test/party_client.py
```diff
@@ -434,3 +434,31 @@ def get_pending_surveys_with_originator_party_id(self, party_id, expected_status
...
+        response = self.client.get(f"/party-api/v1/respondents/{party_id}/password-verification-token")
+        self.assertStatus(response, expected_status)
+        return response.json
+
+    def post_password_verification_token(self, party_id, payload, expected_status=200):
```
LJBabbage: Same again, perhaps I am missing it

test/party_client.py
```diff
@@ -434,3 +434,31 @@ def get_pending_surveys_with_originator_party_id(self, party_id, expected_status
...
+        )
+        self.assertStatus(response, expected_status, response.json)
+
+    def delete_password_verification_token(self, party_id, token):
+        response = self.client.delete(
+            f"/party-api/v1/respondents/{party_id}/password-verification-token/{token}", headers=self.auth_headers
+        )
+        self.assertStatus(response, 200)
+
+    def get_password_reset_counter(self, party_id):
```
LJBabbage: And again

test/party_client.py
```diff
@@ -434,3 +434,31 @@ def get_pending_surveys_with_originator_party_id(self, party_id, expected_status
...
+        response = self.client.post(
+            f"/party-api/v1/respondents/{party_id}/password-verification-token", json=payload, headers=self.auth_headers
+        )
+        self.assertStatus(response, expected_status, response.json)
+
+    def delete_password_verification_token(self, party_id, token):
+        response = self.client.delete(
+            f"/party-api/v1/respondents/{party_id}/password-verification-token/{token}", headers=self.auth_headers
+        )
+        self.assertStatus(response, 200)
```
LJBabbage: Asserting on just 200 is not the best, we should be more explicit, same with the other asserts

test/test_respondent_controller.py
```diff
@@ -1215,6 +1215,30 @@ def test_put_email_verification_uses_case_insensitive_email_query(self):
             account_controller.put_email_verification(token)
             query.assert_called_once_with("test@example.test", db.session())
 
+    def test_email_verification_resets_password_reset(self):
```
LJBabbage: I struggle to understand this test from the name, test_token_removed_on_email_update maybe, something like that. If it's to difficult to explain consider putting in Given, When, Then.

test/test_respondent_controller.py
```diff
@@ -1215,6 +1215,30 @@ def test_put_email_verification_uses_case_insensitive_email_query(self):
...
+        self.populate_with_respondent()
```
LJBabbage: This returns a respondent, so you don't need the line below

LJBabbage: It looks like it might even do a token as well

test/test_respondent_controller.py
```diff
@@ -1215,6 +1215,30 @@ def test_put_email_verification_uses_case_insensitive_email_query(self):
...
+        db_respondent = respondents()[0]
+        token = self.generate_valid_token_from_email(db_respondent.email_address)
+        response = self.put_email_verification(token, 200)
+        db_respondent = respondents()[0]
```
LJBabbage: Not sure why this has been re-assigned to the same thing as above

test/test_respondent_controller.py
```diff
@@ -1215,6 +1215,30 @@ def test_put_email_verification_uses_case_insensitive_email_query(self):
...
+        self.assertIsNone(db_respondent.password_verification_token)
+        self.assertEqual(db_respondent.password_reset_counter, 0)
+        self.assertEqual(response["status"], RespondentStatus.ACTIVE.name)
```
LJBabbage: Checking status, is this needed, there should be another test that does that?

test/test_respondent_controller.py
```diff
@@ -1215,6 +1215,30 @@ def test_put_email_verification_uses_case_insensitive_email_query(self):
...
+
+    def test_email_verification_no_password_reset_token(self):
```
LJBabbage: Similar to the above with comments

test/test_respondent_controller.py
```diff
@@ -1839,31 +1863,31 @@ def test_delete_respondent_mark_for_deletion(self):
...
-        respondent_1 = MockRespondent()
-        respondent_1.attributes(emailAddress="res1@example.com", mark_for_deletion=True)
-        respondent_1 = self.populate_with_respondent(respondent=respondent_1.as_respondent())
-        response = self.delete_user_data_marked_for_deletion()
-        self.assertStatus(response, 204)
-        with self.assertRaises(Exception):
-            self.get_respondent_by_id(respondent.party_uuid)
-        with self.assertRaises(Exception):
-            self.get_respondent_by_email(respondent_1.email_address)
+    # def test_batch_delete_user_data_marked_for_deletion(self):
```
LJBabbage: Why has this been commented out? There should be reason in the code, on the PR. Or if not needed just removed

LJBabbage (changes requested):
This needs a 2nd look. We should also try not to update the Pipfile.lock if not needed, just to limit the amount of things changing outside the scope of this PR. We also should be looking at line coverage when adding code, in this instance that would have helped identify some problems

ras_party/controllers/account_controller.py
arroyoAle: This has been moved up to line 729 right after the email update

ras_party/controllers/account_controller.py
arroyoAle: The second if statement has been removed

ras_party/controllers/account_controller.py
arroyoAle: They are linked but there's no parent function and adding it would require work to frontstage as well and I think is out of the scope of the card

test/party_client.py
arroyoAle: I had added these for completeness but they are now removed

test/party_client.py
arroyoAle: as above

test/party_client.py
arroyoAle: As above

test/party_client.py
arroyoAle: Added expected_status to give more freedom when testing

test/test_respondent_controller.py
arroyoAle: test name has been changed

test/test_respondent_controller.py
arroyoAle: This updates the respondent variable which would otherwise fail the tests as it still contains the verification token and counter is 1

test/test_respondent_controller.py
arroyoAle: This has been removed

test/test_respondent_controller.py
arroyoAle: As above

test/test_respondent_controller.py
arroyoAle: This was just to test locally since it was hanging for a long time locally, has been reverted

test/test_respondent_controller.py
arroyoAle: This has been updated

ras_party/controllers/account_controller.py
```diff
@@ -726,6 +726,10 @@ def put_email_verification(token, tran, session):
 
         if respondent:
             update_verified_email_address(respondent, tran)
+            if not respondent.pending_email_address and respondent.password_verification_token:
```
LJBabbage: This still needs some work, to get to this line we know this is a pending email update. The respondent has received an email which when they clicks updates the email with the pending and then removes the pending_email (line 798) (i.e update_verified_email_address) 

Now because of that and the order this has been put in ```if not respondent.pending_email_address``` becomes redundant as the code underneath will always set respondent.pending_email_address to None, all we need to do is check the verification token existence

test/party_client.py
```diff
@@ -434,3 +434,15 @@ def get_pending_surveys_with_originator_party_id(self, party_id, expected_status
     def delete_pending_surveys_with_batch_no(self, batch_no, expected_status=202):
         response = self.client.delete(f"/party-api/v1/pending-surveys/{batch_no}", headers=self.auth_headers)
         self.assertStatus(response, expected_status)
+
+    def delete_password_verification_token(self, party_id, token, expected_status=200):
+        response = self.client.delete(
+            f"/party-api/v1/respondents/{party_id}/password-verification-token/{token}", headers=self.auth_headers
+        )
+        self.assertStatus(response, expected_status)
```
LJBabbage: This assert is still a bit light and password-verification-token returns ```return jsonify({"message": "Successfully removed token"})``` which we should use

test/party_client.py
```diff
@@ -434,3 +434,15 @@ def get_pending_surveys_with_originator_party_id(self, party_id, expected_status
     def delete_pending_surveys_with_batch_no(self, batch_no, expected_status=202):
         response = self.client.delete(f"/party-api/v1/pending-surveys/{batch_no}", headers=self.auth_headers)
         self.assertStatus(response, expected_status)
+
+    def delete_password_verification_token(self, party_id, token, expected_status=200):
```
LJBabbage: We wouldn't usually have a default var in a function when it is single use. It does make me wonder how we are testing 404 failures, but that is for another story I guess

test/party_client.py
```diff
@@ -434,3 +434,15 @@ def get_pending_surveys_with_originator_party_id(self, party_id, expected_status
...
+        response = self.client.delete(
+            f"/party-api/v1/respondents/{party_id}/password-verification-token/{token}", headers=self.auth_headers
+        )
+        self.assertStatus(response, expected_status)
+
+    def reset_password_reset_counter(self, party_id, expected_status=200):
```
LJBabbage: Both comments above are valid here as well, although ```return jsonify({"message": "Successfully reset counter"})``` this time

test/test_respondent_controller.py
```diff
@@ -1215,6 +1215,26 @@ def test_put_email_verification_uses_case_insensitive_email_query(self):
...
 
+    def test_token_removed_on_email_update(self):
+        respondent = self.populate_with_respondent(respondent=self.mock_respondent_with_pending_email)
+        token = self.generate_valid_token_from_email(respondent.pending_email_address)
+        self.put_email_verification(token, 200)
+        respondent = respondents()[0]
+        self.assertIsNone(respondent.password_verification_token)
+        self.assertEqual(respondent.password_reset_counter, 0)
+
+    def test_email_verification_no_password_reset_token(self):
```
LJBabbage: Do we need this function? does it not have line coverage already?

LJBabbage (changes requested):

test/party_client.py
arroyoAle: For the password token parts the rest is getting tested in the `test_account_controller.py`

ras_party/controllers/account_controller.py
arroyoAle: This has been updated

test/party_client.py
arroyoAle: added functionality to check what message is returned by the api

test/party_client.py
arroyoAle: Added functionality to check what message is returned by the api

test/test_respondent_controller.py
arroyoAle: It's more to check if the section doesn't fall over when there is no verification token as the rest of the tests have a token by default

LJBabbage: Ok, it's not really needed as all you are doing is removing the token and then running the code and making sure it's not there. Not major, but I would removed

LJBabbage (changes requested):
Definitely looking better and party is there as far as I can see, but there is more to do on this story. If you run it into GCP and request a password change and then go in and change the email (by clicking the email etc) the token is removed as expected, however if you then complete the circuit and click the password change email you get a nasty 500 in the code, because of this https://github.com/ONSdigital/ras-frontstage/blob/main/frontstage/views/passwords/reset_password.py#L29. and that is caused as the 404 is just returning https://github.com/ONSdigital/ras-frontstage/blob/main/frontstage/controllers/party_controller.py#L151 The passwords/password-token-not-found.html is probably correct, the code has not been written correctly for the 404

LJBabbage (approved):

PR #406, opened 2023-12-07
Mark for deletion schedule enable disable config

pricem14pc:
# What and why?

We previously suspended all cron jobs involved in account deletion by hardcoding `suspend: true` in the helm template. This PR allows the `party-scheduler-delete-respondents` to be suspended by config per environment as required. We will use this approach for the other three cron jobs in the auth service too. The default is still to suspend the job.

# How to test

Deploy this version of the chart to your env using this associated PR https://github.com/ONSdigital/ras-rm-config/pull/238


SteveScorfield (approved):
Tested and works. Approved

matthew-robinson-ons (approved):
also happy this functions how it should

PR #410, opened 2024-02-27
Change to database exception handling

mfcrocker:
# What and why?
DB exception handling was a little bit of a scattershot approach when implemented because SQLAlchemy's exceptions aren't that good. This is a best attempt at refining it a bit for alerting purposes

# How to test?
- Spin up an environment
- Kill postgres
- Do anything that would call Party
- Check Party logs for the exception

# Jira
RAS-901

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1261,11 +1261,12 @@
         },
         "jinja2": {
             "hashes": [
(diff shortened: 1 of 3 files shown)


SteveScorfield (approved):
Reviewed and tested. Works as expected. PR Approved.

matthew-robinson-ons (approved):
Changes look good and was able to recreate the exception. Might need to resolve the jinja2 vulnerability before merge as it failed the build in other repos.

PR #418, opened 2024-05-09
RAS-1108 Return enrolments based on survey_id and business_id

LJBabbage:
# What and why?
This PRs adds an endpoint to return a list of respondents based on survey_id and business_id
# How to test?
Deploy this branch and either use the acceptance tests or add your own data to get respondents and enrolments. If using the acceptance tests I would recommend adding a 2nd respondent to the same business to prove that 2 results are returned.
Once the data has been added use something like postman to hit the endpoint, don't forget to add the Basic Auth header, it is likely admin/secret. You will find the survey_id and business_id in the enrolment table in party

Screenshot 2024-05-09 at 15 16 57

# Jira


test/test_respondent_controller.py
```diff
@@ -1970,3 +1971,29 @@ def test_respondent_emails_are_case_insensitive(self):
         mock_respondent = self.mock_respondent.copy()
         mock_respondent["emailAddress"] = "A@z.com"
         self.post_to_respondents(payload=mock_respondent, expected_status=409)
+
+    def test_get_enrolments_by_survey_business_id(self):
```
LJBabbage: These tests go against the structure of the others, I wanted to avoid using endpoints to populate the data to test controllers, we really shouldn't be doing that and split the app and endpoints away from the controller tests, it makes the tests cumbersome, difficult to maintain and slow.  It does however look a bit odd mixing them

test/test_respondent_view.py
```diff
@@ -0,0 +1,24 @@
...
+from test.party_client import PartyTestClient
+from test.test_data.default_test_values import (
+    DEFAULT_BUSINESS_UUID,
+    DEFAULT_SURVEY_UUID,
+)
+from test.test_data.mock_respondent import MockRespondent
+from unittest.mock import patch
+
+
+class TestParties(PartyTestClient):
```
LJBabbage, line 11: As above, I have made a test just for the view

test/test_respondent_view.py
```diff
@@ -0,0 +1,31 @@
...
+from test.party_client import PartyTestClient
+from test.test_data.default_test_values import (
+    DEFAULT_BUSINESS_UUID,
+    DEFAULT_SURVEY_UUID,
+)
+from test.test_data.mock_respondent import MockRespondent
+from unittest.mock import patch
+
+
+class TestRespondentView(PartyTestClient):
```
LJBabbage, line 13: As above, I have made a class just for the view, realistically we should separate the tests out, but considering the amount it might just be doing it from now on, and not extending a bad pattern

ras_party/controllers/queries.py
```diff
@@ -582,6 +598,24 @@ def query_enrolment_by_survey_business_respondent(respondent_id, business_id, su
     return response
 
 
+def query_respondents_enrolment_by_survey_and_business_id(survey_id: UUID, business_id: UUID, session: session) -> list:
```
arroyoAle: It feels like this is slightly misnamed since it kind of implies that we are returning respondents while we are just returning respondent_ids

arroyoAle (commented):
All good, tested and correct things are returned just a query on the naming of a function

ras_party/controllers/queries.py
LJBabbage: Good point.

ras_party/models/models.py
```diff
@@ -318,12 +318,14 @@ def to_respondent_dict(self):
             "telephone": self.telephone,
             "status": RespondentStatus(self.status).name,
             "markForDeletion": self.mark_for_deletion,
-            "associations": self._get_business_associations(self.businesses),
```
LJBabbage, line 321: This is causing a bit of extra processing we don't need, so I have created a 2nd function for just this story without it, I will raising a card for all the others to see why and if it is needed

arroyoAle (approved):
Tested and returns expected data and current endpoints work as they should

mfcrocker (approved):
Tested locally, looks good

API.md
```diff
@@ -640,6 +640,44 @@ Returns:
...
+                  "enrolmentStatus":"ENABLED"
+               }
+            ]
+         }
+      ],
+      "password_verification_token":"ImFAei5jb20i.Zj3biA.PAgq3GiYIdfeUywntZS7EDeKIzY",
+      "password_reset_counter":1
+   }
+]
+```
```
pricem14pc: This is not quite right, it's a list of objects each containing the `enrolment_status` and a `respondent` object.
```
    {
        "enrolment_status": "DISABLED",
        "respondent": {
            "emailAddress": "a@z.com",
            "firstName": "A",
            "id": "8010aacc-2858-410e-9c65-c2a542d7589e",
            "lastName": "Z",
            "markForDeletion": false,
            "password_reset_counter": 0,
            "password_verification_token": null,
            "pendingEmailAddress": "b@y.com",
            "sampleUnitType": "BI",
            "status": "ACTIVE",
            "telephone": "01234567890"
        }
    }
```

ras_party/controllers/queries.py
```diff
@@ -582,6 +584,26 @@ def query_enrolment_by_survey_business_respondent(respondent_id, business_id, su
...
+    :param business_id: the uuid of the business
+    :param session: A db session
+    :return: list of respondent ids
+    """
+
+    return (
+        session.query(Respondent, Enrolment.status)
+        .join(Respondent, Respondent.id == Enrolment.respondent_id)
+        .filter(and_(Enrolment.survey_id == survey_id, Enrolment.business_id == business_id))
+        .all()
```
pricem14pc, line 603: Looks ok and returns the expected data. I'm not overly familiar with the best way to form SQLAlchemy queries, or whether chaining a `.join` with a `.filter` is preferable over simply forming the column joins _within_ the filter. It doesn't appear that there's much difference and eyeballing the resulting query is good.

test/test_respondent_view.py
pricem14pc: This looks fine to me and agree with separating this out from the controllers. The only issue I have with doing this is when the views are complex and become too dependent on vast amounts of drifted mocked json. However this feels appropriate.

ras_party/controllers/respondent_controller.py
```diff
@@ -216,6 +221,31 @@ def change_respondent_details(respondent_data, respondent_id, session):
         change_respondent(respondent_data)
 
 
+@with_db_session
+def get_respondents_by_survey_and_business_id(survey_id: UUID, business_id: UUID, session: session) -> list:
+    """
+    Gets a list of Respondents enrolled in a survey for a specified business
```
pricem14pc, line 227: This did stump me for a moment. But it's more a consequence of the business concept of an enrolment, and the fact that an enrolled respondent can separately have a status of `ENABLED` or `DISABLED`. Although we don't test for it explicitly, this implementation correctly returns respondents with _any_ enrolment status (as required by the calling service) 👍

pricem14pc (approved):

PR #427, opened 2024-08-01
RAS-1238 Update to docker compose from docker-compose

arroyoAle:
# What and why?
Updating references to the now removed `docker-compose` in favour of `docker compose`
# How to test?
Only changes to the GHA build
- check that the `Get Dependencies` step works as expected
# Jira
[RAS-1238](https://jira.ons.gov.uk/browse/RAS-1238)


party-compose.sh
arroyoAle, line 1: This script doesn't actually work, it is also completely undocumented and hadn't been touched since 2017 so it might just be worth removing it

anwilkie: I agree, removing it is better. I these commands were a holdover from before we started using Spinnaker/Concourse to set up our services.

matthew-robinson-ons: Also happy for this to be removed as can't see where it is used.

matthew-robinson-ons (approved):
get dependencies looks good with the update

anwilkie (approved):

PR #429, opened 2024-08-21
RAS-1229 Investigate and resolve party service uncaught database exce…

LJBabbage:
# What and why?
The function is purely querying the database and therefore does not need to roll back the session, it also does not need to log out "Rolling back database session due to uncaught exception", but will still raise the NotFound as before
# How to test?
Hit the endpoint http://localhost:8081/party-api/v1/pending-surveys/481ab1a4-463f-410a-a599-70d5b34fd64e and confirm the response is the same before and after the change, but without the logging. Both times it should be 404 
```
{
    "description": "Batch number does not exist"
}
```

# Jira

### DIFF
--- a/_infra/helm/party/Chart.yaml
+++ b/_infra/helm/party/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.5.1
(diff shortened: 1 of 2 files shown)


SteveScorfield (approved):
Works well and code change looks good. Approved.

arroyoAle (approved):
Response is the same and the uncaught error is not longer present

PR #433, opened 2024-09-10
RAS-1228: Bug: Party service workers getting 139 SIGSEGV errors and restarting

SteveScorfield:
# What and why?
There have been a lot of 139 SIGSEGV errors and restarting by the service. The main culprit appears to be the way the calls are being handled and data retrieval of the associations. Due to the complexity of the interconnections between the services, it wasn't possible to separate the calls to associations. There needs to be further investigation into what services are calling the endpoints and whether they actually need the associations.

This code change extracts out the various data dictionaries used by the business and party into one unified dictionary. 

A new endpoint has been added that when called will only not bring back associations. I think this is the way to goto to help minimise the amount of calls being made.

# How to test?

1. Run unit tests 
2. Deploy this along with https://github.com/ONSdigital/response-operations-ui/pull/963 to your dev space
3. Run acceptance tests
4. Run checks to ensure that a CE can be created and linked to a business, that all related information is pulled for a business in from the reporting unit page (via the reporting search page), that the respondent data is being pulled correctly, messages can be sent and read correctly in rOps
5. In Frontstage, just a general regression check to check everything is working correctly. 

# Jira
https://jira.ons.gov.uk/browse/RAS-1228

### DIFF
--- a/ras_party/models/model_functions.py
+++ b/ras_party/models/model_functions.py
@@ -1,0 +1,40 @@
+import logging
+
+import structlog
+from werkzeug.exceptions import BadRequest
+
+from ras_party.models import models
+
+logger = structlog.wrap_logger(logging.getLogger(__name__))
+
+
+def get_respondents_associations(respondents):
+    associations = []
+    for business_respondent in respondents:
+        respondent_dict = {
+            "partyId": business_respondent.respondent.party_uuid,
+            "businessRespondentStatus": business_respondent.respondent.status.name,
+        }
+        enrolments = business_respondent.enrolment
+        respondent_dict["enrolments"] = []
+        for enrolment in enrolments:
+            enrolments_dict = {
+                "surveyId": enrolment.survey_id,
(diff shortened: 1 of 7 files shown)


ras_party/unified_buisness_party_functions.py
SteveScorfield, line 1: Not sure I like the naming of this. But I couldn't think what would be a good name for this.

ras_party/models/model_functions.py
SteveScorfield, line 1: Due to the retrieval of the enrolment DTO, I had to extract these functions out of the unified_business_party_functions.py

ras_party/views/business_view.py
```diff
@@ -79,6 +79,13 @@ def get_business_by_ref(ref):
     return jsonify(business)
 
 
+@business_view.route("/businesses/ref/reporting-unit-only/<ref>", methods=["GET"])
```
SteveScorfield, line 82: Not keen on the naming of this endpoint. Any ideas?

PR #434, opened 2024-09-11
Refactor email token decoding

matthew-robinson-ons:
# What and why?
When put_email_verification or verify_token encounter an exception during the decoding of the email token, it is raised within the session and causes incorrect logging and session rollbacks. 

This PR moves the decoding of the email token outside of the session context for both functions and reduced duplicate code within the functions. 

Flask-cors also had a vulnerability which has been updated on this pr. 
# How to test?
Ensure email verification still works correctly when updating or registering an email and the calls to the party service no longer contain session rollbacks or incorrect logging when an exception is thrown. 
# Jira
https://jira.ons.gov.uk/secure/RapidBoard.jspa?rapidView=1493&projectKey=RAS&view=detail&selectedIssue=RAS-1262

### DIFF
--- a/ras_party/controllers/account_controller.py
+++ b/ras_party/controllers/account_controller.py
@@ -408,16 +408,14 @@ def decode_token(token):
 
 
 @with_query_only_db_session
-def verify_token(token, session):
+def verify_respondent_by_email(email, session):
     """
     Verify the email token and check if that respondent exists.
-    :param token: The email token for verification
+    :param email: The email token for verification
     :param session: Database session
     :return: Verification response
     """
-
-    email_address = decode_token(token)
-    respondent = query_respondent_by_email(email_address, session)
+    respondent = query_respondent_by_email(email, session)
     if not respondent:
         logger.info("Respondent with Email from token does not exist")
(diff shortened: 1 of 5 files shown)


ras_party/controllers/account_controller.py
```diff
@@ -389,18 +389,34 @@ def _send_account_email_changed_notification(email_address, new_email_address, r
...
+@with_query_only_db_session
+def verify_token(token, session):
+    """
+    Verify the email token and check if that respondent exists.
+    :param token: The email token for verification
+    :param session: Database session
+    :return: Verification response
+    """
 
+    email_address = decode_token(token)
```
LJBabbage: You're still inside the decorator and have just refactored the function, put_email_verification should just be getting an email, the token decode should be done outside session scope

matthew-robinson-ons: Have moved the decode outside of the function now. Finding it difficult to test this change but believe the recent changes have solved it.

anwilkie (approved):
Looks good to me after testing.

LJBabbage (approved):

PR #435, opened 2024-09-13
RAS-1289 Update the way claims is determined in the party service

LJBabbage:
# What and why?
This PR updates how claims are processed and introduces survey_id in the check. This is part of a larger piece of reducing the amount of 139 crashes, at present get_respondent_by_id is doing more that what it should and by calling to_respondent_with_associations_dict() causing considerable overhead leading to timeouts. I can't remove get_respondent_by_id or the dict it is generating, but I can move the functionality to a cleaner function which just returns the repondent and then make enrolment checks cleaner and far more efficient

# How to test?
Test in conjunction with https://github.com/ONSdigital/ras-secure-message/pull/428, trying sending a new email via the survey help route, the account help route and replying to an existing thread. Reply and create a new message in ROPs as well
# Jira

### DIFF
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -13,6 +13,7 @@ from ras_party.controllers.account_controller import (
 )
 from ras_party.controllers.notify_gateway import NotifyGateway
 from ras_party.controllers.queries import (
+    query_enrolment_by_survey_business_respondent,
     query_respondent_by_email,
     query_respondent_by_names_and_emails,
     query_respondent_by_party_uuid,
@@ -25,6 +26,7 @@ from ras_party.models.models import (
     Enrolment,
     PendingEnrolment,
     Respondent,
+    RespondentStatus,
 )
 from ras_party.support.session_decorator import (
     with_db_session,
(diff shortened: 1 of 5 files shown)


ras_party/controllers/respondent_controller.py
```diff
@@ -69,15 +71,25 @@ def get_respondents_by_name_and_email(first_name, last_name, email, page, limit,
     }
 
 
+def get_respondent_by_party_uuid(party_uuid: UUID, session: session) -> Respondent:
```
LJBabbage, line 74: At the moment this is quite fluid, I might wrap it in a session at some stage, but decided not to at present.

ras_party/views/respondent_view.py
```diff
@@ -113,16 +113,14 @@ def validate_respondent_claim():
     in client services. There is an argument to use a 403 on invalid claims , but that should be a status on the
     resource not the state of the returned data and so that's stretching the use of http status codes somewhat
     """
-    respondent_id = request.args.get("respondent_id", default="").strip()
-    business_id = request.args.get("business_id", default="").strip()
+    party_uuid = request.args.get("respondent_id")
```
LJBabbage, line 116: The var names passed from SM and here are a muddle, I don't intend fixing them outside the flow of this story

ras_party/views/respondent_view.py
```diff
@@ -113,16 +113,14 @@ def validate_respondent_claim():
...
+    business_id = request.args.get("business_id")
+    survey_id = request.args.get("survey_id")
```
mfcrocker, line 118: Given you've added the survey ID to the API, could you please update openapi.md?

mfcrocker (approved):
Looks good, one bit of housekeeping

anwilkie (approved):

PR #437, opened 2024-10-03
RAS-1290 Add clean get respondent by party_id endpoint to party service

LJBabbage:
# What and why?
This Pr adds a new endpoint to get respondents by party_id. This card is part of bigger piece to removed the business associations currently returning from the respondents by id endpoint
# How to test?
Deploy this PR and run the acceptance tests (or add a respondent) make sure when you call the endpoint you get the right response. Don't forget to include basic auth details 
Screenshot 2024-10-03 at 11 37 19

# Jira

### DIFF
--- a/ras_party/models/models.py
+++ b/ras_party/models/models.py
@@ -322,6 +322,22 @@ class Respondent(Base):
             "password_reset_counter": self.password_reset_counter,
         }
 
+    def to_dict(self) -> dict:
+        return {
+            "id": self.id,
+            "party_uuid": self.party_uuid,
+            "status": RespondentStatus(self.status).name,
+            "email_address": self.email_address,
+            "pending_email_address": self.pending_email_address,
+            "first_name": self.first_name,
+            "last_name": self.last_name,
+            "telephone": self.telephone,
+            "created_on": self.created_on,
+            "mark_for_deletion": self.mark_for_deletion,
+            "password_verification_token": self.password_verification_token,
+            "password_reset_counter": self.password_reset_counter,
+        }
+
     def to_respondent_with_associations_dict(self):
         respondent_dict = self.to_respondent_dict()
         respondent_dict["associations"] = self._get_business_associations(self.businesses)
--- a/ras_party/views/respondent_view.py
+++ b/ras_party/views/respondent_view.py
@@ -77,6 +77,18 @@ def _validate_get_respondent_params(ids, first_name, last_name, email):
                 raise BadRequest(f"'{party_id}' is not a valid UUID format for property 'id'")
 
 
+@respondent_view.route("/respondents/party_id/<party_id>", methods=["GET"])
+def get_respondent_by_party_id(party_id: UUID) -> Response:
+    if not is_valid_uuid4(party_id):
+        return make_response("party_id is not UUID", 400)
+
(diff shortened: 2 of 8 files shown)


ras_party/models/models.py
```diff
@@ -322,6 +322,22 @@ def to_respondent_dict(self):
             "password_reset_counter": self.password_reset_counter,
         }
 
+    def to_dict(self) -> dict:
```
LJBabbage, line 325: This clearly isn't ideal, as we have a to_respondent_dict as well. The problem with that one is not only assign party_id to id, but also the keys are a mix of camelCase and snake, it also returns an odd UNIT_TYPE which is a constant and therefore not persisted in the db. Naturally I can't just change to_respondent_dict as it will break anything calling to it.

However as this new endpoint is a migration piece, to_respondent_dict will eventually be removed.

ras_party/views/respondent_view.py
```diff
@@ -77,6 +77,18 @@ def _validate_get_respondent_params(ids, first_name, last_name, email):
                 raise BadRequest(f"'{party_id}' is not a valid UUID format for property 'id'")
 
 
+@respondent_view.route("/respondents/party_id/<party_id>", methods=["GET"])
+def get_respondent_by_party_id(party_id: UUID) -> Response:
+    if not is_valid_uuid4(party_id):
```
LJBabbage, line 82: The validation should not be inside the decorator as nearly everywhere else. Any exception will get raised in the decorator into a session that we don't want or need if it's not valid.

ras_party/views/respondent_view.py
```diff
@@ -77,6 +77,18 @@ def _validate_get_respondent_params(ids, first_name, last_name, email):
...
+        return make_response("party_id is not UUID", 400)
+
+    respondent = respondent_controller.get_respondent_by_party_id(party_id)
+    if respondent:
```
LJBabbage, line 86: I only want to deal with the object until it gets returned, why the conversion is here.

ras_party/views/respondent_view.py
```diff
@@ -77,6 +77,18 @@ def _validate_get_respondent_params(ids, first_name, last_name, email):
...
+        return make_response(respondent.to_dict(), 200)
+
+    return make_response(f"respondent not found for party_id {party_id}", 404)
```
LJBabbage, line 89: A none response from sql is perfectly fine in the controller, I only want to start using 404 response when it is returning, using werkzeug.exceptions http responses in a generic controller is slightly clumsy and should be avoided if possible. This is however a work in progress as this will need to be integrated with other uses of the code at some stage

arroyoAle (approved):
Looks good and tested - everything returns properly and error handling works

pricem14pc:
As intended, SQLAlchemy query is nice and clean 👍 
```
SELECT 
partysvc.respondent.id AS partysvc_respondent_id,
 ...
 ...
partysvc.respondent.password_reset_counter AS partysvc_respondent_password_reset_counter
FROM partysvc.respondent
WHERE partysvc.respondent.party_uuid = 'b90478a4-0eb6-4f8c-865a-d5f2df7d09ff'::UUID
LIMIT 1
```

pricem14pc (approved):
Is it worth adding in the openapi spec that the existing overloaded method is deprecated?

PR #443, opened 2024-11-14
RAS-1350 Create enrolment end point for party

LJBabbage:
# What and why?
This PR adds in an endpoint to retrieve enrolments, I have intentionally made it wider than it probably needs. The respondent-details endpoint which is causing us all the problems and unnecessarily builds enrolments in a very inefficient way is used indirectly in over 50 places across our microservices. Although I have looked at everyone, I can't confidently say exactly the requirements of these 2 end points until I start working individually on each, so I have made both more inclusive with the aim to harden afterwards. 

N.B there are also now places I can refactor using this approach in this repo, things like the claim end point can be removed, plus other functions I can plug into the query, I don't want to over stretch here. This is purely about getting 2 endpoints (this and the respondent one) which are clean and efficient and then migrating over in a controlled way

# How to test?
Deploy this PR and put some data into the system (acceptance tests are fine) Then hit the endpoint with different payloads and see what you get back, don't forget your basic auth (Username: admin, Passord: secret)
Screenshot 2024-11-15 at 10 10 05

# Jira

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,13 +4,13 @@ build:
 	pipenv install --dev
 
 lint:
-	pipenv check -i 42194 -i 70612 -i 70624 -i 72731
+	pipenv check -i 42194 -i 70612 -i 70624 -i 72731 -i 73655
 	pipenv run isort .
 	pipenv run black --line-length 120 .
 	pipenv run flake8 --exclude=./scripts
 
 lint-check:
-	pipenv check -i 42194 -i 70612 -i 70624 -i 72731
+	pipenv check -i 42194 -i 70612 -i 70624 -i 72731 -i 73655
 	pipenv run isort . --check-only
 	pipenv run black --line-length 120 --check .
 	pipenv run flake8 --exclude=./scripts
--- a/Makefile
+++ b/Makefile
@@ -4,13 +4,13 @@ build:
 	pipenv install --dev
 
 lint:
-	pipenv check -i 42194 -i 70612 -i 70624 -i 72731 -i 73655
+	pipenv check -i 42194 -i 70612 -i 70624 -i 72731
 	pipenv run isort .
 	pipenv run black --line-length 120 .
 	pipenv run flake8 --exclude=./scripts
 
 lint-check:
-	pipenv check -i 42194 -i 70612 -i 70624 -i 72731 -i 73655
+	pipenv check -i 42194 -i 70612 -i 70624 -i 72731
 	pipenv run isort . --check-only
 	pipenv run black --line-length 120 --check .
 	pipenv run flake8 --exclude=./scripts
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -823,12 +823,20 @@ paths:
           description: The password reset counter has been reset
         404:
           description: The respondent does not exist
-  /enrolments:
+  /enrolments/respondent/{party_uuid}:
     get:
       tags:
         - enrolments
       summary: get enrolment details
       description: returns a list of all enrolments that match given parameters
+      parameters:
+        - name: party_uuid
+          in: path
+          required: true
+          description: The UUID of the respondent
+          schema:
+            type: string
+            format: uuid
       responses:
         200:
           description: list of dict enrolments
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -11,6 +11,8 @@ tags:
     description: Respondent endpoints
   - name: businesses
     description: Business endpoints
+  - name: enrolments
+    description: enrolments endpoints
   - name: info
     description: Information endpoints
   - name: misc
@@ -821,6 +823,37 @@ paths:
           description: The password reset counter has been reset
         404:
           description: The respondent does not exist
+  /enrolments:
+    get:
+      tags:
+        - enrolments
+      summary: get enrolment details
+      description: returns a list of all enrolments that match given parameters
+      responses:
+        200:
+          description: list of dict enrolments
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: object
+                  properties:
+                    business_id:
+                      type: string
+                      format: uuid
+                    respondent_id:
+                      type: integer
+                    status:
+                      type: string
+                      example: ENABLED
+                    survey_id:
+                      type: string
+                      format: uuid
+        400:
+          description: Missing of malformed parameters
+        404:
+          description: Respondent doesn't exist
   /batch/respondents:
     delete:
       tags:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -850,8 +850,6 @@ paths:
                     business_id:
                       type: string
                       format: uuid
-                    respondent_id:
-                      type: integer
                     status:
                       type: string
                       example: ENABLED
--- a/ras_party/controllers/enrolments_controller.py
+++ b/ras_party/controllers/enrolments_controller.py
@@ -6,8 +6,8 @@ from flask import session
 from sqlalchemy.orm.exc import NoResultFound
 
 from ras_party.controllers.queries import (
-    query_enrolments_by_parameters,
     query_respondent_by_party_uuid,
+    query_respondent_enrolments,
 )
 from ras_party.models.models import Enrolment
 from ras_party.support.session_decorator import with_query_only_db_session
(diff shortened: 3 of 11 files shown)


Makefile
```diff
@@ -4,13 +4,13 @@ build:
 	pipenv install --dev
 
 lint:
-	pipenv check -i 42194 -i 70612 -i 70624 -i 72731
+	pipenv check -i 42194 -i 70612 -i 70624 -i 72731 -i 73655
```
LJBabbage: I will raise a card for this, just didn't want to fix it here

ras_party/models/models.py
```diff
@@ -390,6 +390,14 @@ class Enrolment(Base):
         ),
     )
 
+    def to_dict(self) -> dict:
+        return {
+            "business_id": self.business_id,
+            "respondent_id": self.respondent_id,
```
LJBabbage: Very debatable if status of respondent_id should be passed back, respondent_id is the int in enrolment rather than the UUID so passing that back might be pointless

ras_party/views/enrolments_view.py
```diff
@@ -0,0 +1,60 @@
...
+@enrolments_view.route("/enrolments", methods=["GET"])
+def get_enrolments() -> Response:
+    json = request.get_json()
+    party_uuid = json.get("party_uuid")
+    business_id = json.get("business_id")
+    survey_id = json.get("survey_id")
+    status = json.get("status")
+
+    if not (party_uuid or business_id or survey_id):
+        logger.error("No parameters passed to get_enrolments")
```
LJBabbage: I will change this one with any review comments, I don't like the error message, I will make it more payload specific

test/test_enrolments_controller.py
```diff
@@ -0,0 +1,138 @@
...
+                "business": "75d9af56-1225-4d43-b41d-1199f5f89daa",
+                "survey_id": "9200d295-9d6e-41fe-b541-747ae67a279f",
+                "status": EnrolmentStatus.DISABLED,
+            },
+        ],
+    },
+]
+
+
+class TestEnrolments(PartyTestClient):
```
LJBabbage, line 50: There are many different other combinations I could have done, but I don't think they are needed, so covered each main one individually, a combination and then the errors that could be raised

ras_party/controllers/enrolments_controller.py
```diff
@@ -0,0 +1,33 @@
...
+
+@with_query_only_db_session
+def enrolments_by_parameters(
+    session: session, party_uuid: UUID = None, business_id: UUID = None, survey_id: UUID = None, status: int = None
+) -> list[Enrolment]:
+    """
+    returns a list of Enrolments based on provided parameters
+    """
+
+    if party_uuid:
```
LJBabbage: I don't want to join the tables, so decided to do it this way. Enrolment does not have the respondent uuid which is being passed in virtually every use I have seen

ras_party/controllers/queries.py
```diff
@@ -643,3 +643,23 @@ def count_enrolment_by_survey_business(business_id, survey_id, session):
...
+    if respondent_id:
+        conditions.append(Enrolment.respondent_id == respondent_id)
+    if business_id:
+        conditions.append(Enrolment.business_id == business_id)
+    if survey_id:
+        conditions.append(Enrolment.survey_id == survey_id)
+    if status:
+        conditions.append(Enrolment.status == status)
+
+    return session.query(Enrolment).filter(and_(*conditions)).all()
```
LJBabbage: Could this return to much and need restricting with a count, maybe. It depends on how it is used. Nearly ever use I have seen is respondent/business driven which means this should be fine. Calling to it with just survey_id could produce a large result set, but I don't believe we every do that. As said above this could be a  place I might harden after I start plugging it in

arroyoAle (approved):
All looks fine, everything works, returns expected values, and errors get handled. Only thing is when a UUID is malformed it prints out a large stacktrace to the logs, not really connected to this card but more the session handling since it is within a log

ras_party/views/enrolments_view.py
arroyoAle: I guess it gets the point across but maybe something that tells us what we could pass?

LJBabbage:
> All looks fine, everything works, returns expected values, and errors get handled. Only thing is when a UUID is malformed it prints out a large stacktrace to the logs, not really connected to this card but more the session handling since it is within a log

Yep all of them have it  (exc_info=True) In this scenario it doesn't matter hugely, but for anything with emails or sensitive information it could be punched out into the logs

pricem14pc:
Fat fingers sorry @LJBabbage 🖐️

ras_party/models/models.py
pricem14pc: As discussed off-line, I don't think the API should return the internal PK of the database record (i.e. in this case the PK of the respondent record).

However, I acknowledge that if this is omitted multiple respondent enrollments for the same business/survey will appear as duplications with no differentiation between each. It's only the inclusion of the `respondent_id` that's providing uniqueness as an enrolment is scoped to a respondent. The ideal solution for this API would be to return the respondent party UUID instead of the db PK (even if it's not needed for the current use cases being addressed).

At the very least we should document/capture that we are intending a future extension of the `/enrolments` API to cover wider use cases to that may require the UUID of the respondent to be included.

This new approach is not a direct replacement for the list of `"associations": [...]` in the current API, it is a step towards breaking up the bloated requests/responses that are causing performance issues in high frequency pages.

ras_party/controllers/queries.py
pricem14pc: As we can see from other APIs used/extended over the years, there is a danger that other developers misuse or don't recognise the implications of calling this in future solutions. Again, at the very least should caveat its usage in its definition.

ras_party/controllers/enrolments_controller.py
pricem14pc: As discussed off-line and commented below, for the use cases this PR is intending to address/improve this is fine. However, we should either tweak the `/enrolments` API so the respondent uuid is not optional (e.g. `/enrolments/<RESPONDENT_UUID>` or join the tables to return the uuid if it's not included as a parameter.

The former is probably fine for Frontstage use cases, but the later may be needed for Rops use cases.

ras_party/controllers/queries.py
LJBabbage: We do need to trust developers and the review process even if it has broken in the past, otherwise we are struggling. I will harden it to be just respondents

openapi.yaml
```diff
@@ -821,6 +823,45 @@ paths:
...
+              schema:
+                type: array
+                items:
+                  type: object
+                  properties:
+                    business_id:
+                      type: string
+                      format: uuid
+                    respondent_id:
+                      type: integer
```
pricem14pc: I don't think we're returning this any more? I see
```
[
    {
        "business_id": "7096b277-6f96-4d73-b872-1837f24164fc",
        "status": "ENABLED",
        "survey_id": "0b1f8376-28e9-4884-bea5-acf9d709464e"
    }
]
```

LJBabbage: You're right i missed removing the respondent_id

test/test_enrolments_controller.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def test_get_enrolments_party_id_and_business_id_and_survey_id(self):
+        enrolments = respondent_enrolments(
+            party_uuid="b6f9d6e8-b840-4c95-a6ce-9ef145dd1f85",
+            business_id="75d9af56-1225-4d43-b41d-1199f5f89daa",
+            survey_id="9200d295-9d6e-41fe-b541-747ae67a279f",
+        )
+
+        self.assertEqual(len(enrolments), 1)
+        self.assertIn(str(enrolments[0].respondent_id), "b6f9d6e8-b840-4c95-a6ce-9ef145dd1f85")
```
pricem14pc: This assertion is inadvertently passing. The `enrolments[0].respondent_id` correctly yields the integer `1` (i.e. the db sequence key for the respondent, but we're then asserting that the string value `1` is in the UUID string `b6f9d6e8-b840-4c95-a6ce-9ef145dd1f85` which it is.

These assertions could all be `assertEqual` with the respondent_id being changed to
```
self.assertIn(str(enrolments[0].respondent_id), "1")
```

pricem14pc (commented):
Minor changes only

test/test_enrolments_controller.py
LJBabbage: That is a cracking spot, and you're right it should be assertEqual for everything. I must of just copied and pasted after initially getting it wrong or did it differently by using in first

pricem14pc (approved):

PR #445, opened 2024-11-18
update python version (w/ datetime bug)

anwilkie:
# What and why?
Updating python version to 3.12.

# How to test?
* Install local pipenv environment. (You may have a file called python-version or something along those lines on your local branch, which you will need to delete; it's an unneeded file leftover from earlier versions of the repo that isn't part of this branch or on main)
* Deploy
* Run acceptance tests
* Go onto r-ops and hit party endpoints

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-1351)


matthew-robinson-ons (changes requested):
I am requiring more pipfile.lock changes to run the app locally. If i remove the the pipenv and install a new one on python 3.12 the markers need updating within the pipfile. There are also deprecation warnings now we have moved to 3.12 for `datetime.datetime.utcnow() ` we should have a card to address this as we use it in a lot of places.

LJBabbage (changes requested):
I got the same as Matthew, it called me out for not have an up to date lock file using the one on this PR, it also didn't like isort for some reason. To answer Matthew depreciation question, unless large it should be fixed on this PR as it is part of the update process. 
Better testing instructions would also be appreciated, I regularly forget how to update my pipenv and get the ```Specifier python_version does not match 3.12 (3.11).```
This kind of stuff is always good to have on a PR as you must of had to do it yourself

anwilkie:
I've tried changing the instances of `utcnow` to `now(timezone.uct)` in order to fix the deprecation warning ([as recommended online](https://docs.python.org/3/library/datetime.html#datetime.datetime.now)), but it introduces a strange bug where the timestamps generated by the service become inaccurate.

To replicate, deploy this PR, run acceptance tests, and check the `created_on` fields in the party database tables, then run the acceptance tests again. You'll notice that the enrolments/accounts are recreated, but the timestamps stay "stuck" on the same hour and minute as the previous iteration, and sometimes the seconds change. This also occurs if you manually create an account yourself.

I'm puzzled as to what's causing this.

PR #446, opened 2024-11-21
Remove respondent dict and update test

LJBabbage:
# What and why?
Whilst writing the new respondent end point a new dict was introduced. It was to not only sort out the structure, but to make the json more consistent, in reflection it is too aspirational to also fix at the same time, and the old one should be used

# How to test?
The good news is the only endpoint this is used by has no use by any other services yet, and that endpoint is now using a well tested dict. This basically doesn't need anything more than an eye ball
# Jira


anwilkie (approved):

SteveScorfield (approved):
Looks fine. Approved

PR #447, opened 2024-11-22
python 3.12 upgrade

anwilkie:
# What and why?
A redo of the python 3.12 upgrade. There is a known deprecation of the `datetime.datetime.utcnow` function as of this version, but trying to fix it resulted in a bug involving the timestamps being introduced. That issue will be resolved as part of a separate card.

# How to test?
* Do a fresh pipenv install (N.B. You may have a file called `python-version` or something along those lines on your local branch. Just delete it--it's a file leftover from earlier versions of the repo and it's not present on this branch or on main)
* Deploy
* Run acceptance tests and check that they pass.
* Do things related to the party service on frontstage/r-ops, such as creating a new user/enrolment, editing respondent data, and so on. They should still work.

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-1351)


matthew-robinson-ons (approved):
Large pipfile update, though can be necessary for a python upgrade. Functions locally and when deployed to GCP. The deprecation warnings are present but if problematic seems appropriate to have them on a separate card.

LJBabbage (commented):
My isort still wants updating, not sure if that is isolated

anwilkie:
I can't see anything wrong with isort. I've updated it to the most recent version as of the time of writing.

mfcrocker (approved):

PR #448, opened 2024-12-06
RAS-1378 Fix datetime deprecation warning on party service

arroyoAle:
# What and why?
Delegates the created_on time retrieval to the built in database method (which saves it in UTC without timezone information - as we have been doing with datetime anyway). This solves the issue we were seeing with using the "new" `datetime.now(UTC)` function replacing the deprecated `datetime.utcnow()` function, which was using the time checked at app startup in the models and not evaluating on each column creation. The instances in the pending survey deletion cron job are evaluated on each request and therefore work as they should.
# How to test?
- deploy to your env
- run the acceptance tests
- check the created on time for the new party 
- create a new account in frontstage
- check the time is not set to almost the same but the correct time the account was created
- share the survey to the other account you have
- check that there is a pending survey share in the `pending_surveys` table
- run `kubectl edit deployments party` and edit the `EMAIL_TOKEN_EXPIRY` to a more reasonable number from 3 days
  - or edit it through the gcloud user interface
- save the changes
- check that after the cron runs the pending survey entry has been deleted from the table
# Jira
[RAS-1378](https://jira.ons.gov.uk/browse/RAS-1378)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @ONSdigital/sdc-rmras
--- a/_infra/helm/party/Chart.yaml
+++ b/_infra/helm/party/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.5.12
+version: 2.5.13
 
 # This is the version number of the application being deployed. This version number should be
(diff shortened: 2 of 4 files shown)


LJBabbage (approved):
Worked fine for me. I just changed the date in the db for pending survey, seemed easier

matthew-robinson-ons (approved):
Works fine, times all look accurate and pending survey is deleted as expected.

PR #449, opened 2024-12-17
RAS-1400 Extend party enrolment to include business and survey details

LJBabbage:
# What and why?
This PR updates the enrolment information endpoint, instead of just posting busines_ids and survey_ids it provides all the bits that are needed. It also adds a new endpoint just to checkout enrolment
# How to test?
Deploy this card as well https://github.com/ONSdigital/ras-frontstage/pull/1016. Add some enrolment data to your account, acceptance tests could be used, but you might want to add another IAC as well. If it logs in you are good. Also check you can access a eq, download and upload a seft. Equally you can just call the endpoints on this service

N.B there is a chart addition so you will need to deploy this via helm or provide the survey_url env. (i.e helm install party ./_infra/helm/party) don't forget if you already have a party service deployed via spinnaker you will need to kill the current deployments, service etc

```kubectl delete deployment party --namespace <your namespace>

### DIFF
--- a/ras_party/controllers/enrolments_controller.py
+++ b/ras_party/controllers/enrolments_controller.py
@@ -6,10 +6,12 @@ from flask import session
 from sqlalchemy.orm.exc import NoResultFound
 
 from ras_party.controllers.queries import (
+    query_enrolment_by_survey_business_respondent,
     query_respondent_by_party_uuid,
     query_respondent_enrolments,
 )
-from ras_party.models.models import Enrolment
+from ras_party.controllers.survey_controller import get_surveys_details
+from ras_party.models.models import Enrolment, RespondentStatus
 from ras_party.support.session_decorator import with_query_only_db_session
 
 logger = structlog.wrap_logger(logging.getLogger(__name__))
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: ras_party/controllers/enrolments_controller.py [hunk 1]
```diff
@@ -20,11 +22,50 @@ def respondent_enrolments(
...
+                        "ref": surveys_details.get(survey_id)["ref"],
+                    },
+                }
+            )
+        )
+    return respondents_enrolled
+
+
+@with_query_only_db_session
+def is_respondent_enrolled(party_uuid: UUID, business_id: UUID, survey_id: UUID, session: session) -> bool:
```
LJBabbage, line 64: I've moved this, with the aim to shift anything associated to enrolments to the enrolments controller

ras_party/controllers/survey_controller.py
```diff
@@ -0,0 +1,31 @@
...
+import structlog
+from flask import current_app
+from requests.exceptions import ConnectionError, HTTPError, Timeout
+
+from ras_party.exceptions import ServiceUnavailableException
+
+logger = structlog.wrap_logger(logging.getLogger(__name__))
+
+
+def get_surveys_details() -> dict:
```
LJBabbage, line 13: In the future, I would like to make this a single app call, cache etc. At the moment this is fine

ras_party/views/enrolments_view.py
```diff
@@ -58,4 +61,17 @@ def get_respondent_enrolments(party_uuid: UUID) -> Response:
...
         return BadRequest()
 
-    return make_response([enrolment.to_dict() for enrolment in enrolments], 200)
+    return make_response(enrolments, 200)
+
+
+@enrolments_view.route(
+    "/is_respondent_enrolled/<party_uuid>/business_id/<business_id>/survey_id/<survey_id>", methods=["GET"]
+)
+def get_is_respondent_enrolled(party_uuid: UUID, business_id: UUID, survey_id: UUID) -> Response:
```
LJBabbage, line 70: There is a similar endpoint called validate_respondent_claim, I am not a fan and will look to use just this one in the future. The only reason I haven't done it now is that is used by secure message, and rather than expanding the scope and having to update a 3rd repo, I would rather do a small change there after this PR is done

matthew-robinson-ons (approved):
Functions properly, no issues with the suggested code changes

SteveScorfield (approved):
Changes look fine and tests all okay. Approved

PR #450, opened 2025-01-06
RAS-1400 use native sql

LJBabbage:
# What and why?
The code in the model https://github.com/ONSdigital/ras-party/blob/main/ras_party/models/models.py#L40 will force join the attributes table, due to how we create business attributes for every individual sample entry, means there are literally thousands for some businesses. Because of that if we attempt any join using the Business model it will overwork and produce an excessive performance hitting amount of joins. This PR sides steps the model and creates a pure SQL of exactly what we want

# How to test?
Create enrolments for different CE's and make sure when you log in the correct information is displayed on the todo list. You can use acceptance-tests, but I would add another enrolled to them and possible another CE
# Jira

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -839,23 +839,11 @@ paths:
             format: uuid
       responses:
         200:
-          description: list of dict enrolments
+          description: dict of respondent enrolments and associated business and survey details
           content:
             application/json:
               schema:
-                type: array
-                items:
-                  type: object
-                  properties:
-                    business_id:
-                      type: string
-                      format: uuid
-                    status:
-                      type: string
-                      example: ENABLED
-                    survey_id:
-                      type: string
-                      format: uuid
+                $ref: '#/components/schemas/RespondentEnrolments'
         400:
           description: Missing of malformed parameters
         404:
@@ -1727,3 +1715,31 @@ components:
           type: string
           format: date-time
           example: 2020-01-01T00:00:00Z
+    RespondentEnrolments:
+      type: object
+      properties:
+        enrolment_status:
+          type: string
+          enum: [PENDING, ENABLED, DISABLED, SUSPENDED]
+        business_details:
+          type: object
+          properties:
+            id:
+              format: uuid
+            name:
+              type: string
+            trading_as:
+              type: string
+            ref:
+              type: string
+        survey_details:
+          type: object
+          properties:
+            id:
+              format: uuid
+            long_name:
+              type: string
+            short_name:
+              type: string
+            ref:
+              type: string
--- a/ras_party/controllers/queries.py
+++ b/ras_party/controllers/queries.py
@@ -664,9 +664,9 @@ def query_respondent_enrolments(
 
     return session.execute(
         text(
-            f"SELECT * FROM partysvc.enrolment "
+            f"SELECT business_id, status, survey_id, business_ref, attributes from partysvc.enrolment "
             f"inner join partysvc.business on partysvc.business.party_uuid = partysvc.enrolment.business_id, "
-            f"LATERAL (SELECT * FROM partysvc.business_attributes "
+            f"LATERAL (SELECT attributes FROM partysvc.business_attributes "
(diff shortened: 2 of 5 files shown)


SteveScorfield (approved):
I have tested the code and works as expected. The acceptance tests pass. I like this approach. It's more streamlined. Approved.

ras_party/controllers/queries.py
```diff
@@ -649,21 +650,24 @@ def query_respondent_enrolments(
...
-        session.query(Enrolment, Business.business_ref, BusinessAttributes.attributes)
-        .join(Business, Business.party_uuid == Enrolment.business_id)
-        .join(BusinessAttributes, BusinessAttributes.business_id == Enrolment.business_id)
-        .filter(and_(Enrolment.respondent_id == respondent_id, *additional_conditions))
-        .all()
+        where_clause += f" and partysvc.enrolment.status='{status}'"
+
+    return session.execute(
+        text(
+            f"SELECT * FROM partysvc.enrolment "
```
pricem14pc: This would benefit from explicitly returning the table fields we want to map to the Python object rather than wilcarding all the fields.

openapi.yaml
```diff
@@ -839,23 +839,11 @@ paths:
             format: uuid
```
pricem14pc, line 839: I think the correct request definition for a required empty filter object of `{ }` with optional parameters would be this
```
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: object
              minProperties: 0
              properties:
                business_id:
                  type: string
                  format: uuid
                survey_id:
                  type: string
                  format: uuid
                status:
                  type: string
                  enum: ["ENABLED", "DISABLED"]
```

ras_party/controllers/queries.py
```diff
@@ -649,21 +650,24 @@ def query_respondent_enrolments(
...
-        .join(BusinessAttributes, BusinessAttributes.business_id == Enrolment.business_id)
-        .filter(and_(Enrolment.respondent_id == respondent_id, *additional_conditions))
-        .all()
+        where_clause += f" and partysvc.enrolment.status='{status}'"
+
+    return session.execute(
+        text(
+            f"SELECT business_id, status, survey_id, business_ref, attributes from partysvc.enrolment "
+            f"inner join partysvc.business on partysvc.business.party_uuid = partysvc.enrolment.business_id, "
+            f"LATERAL (SELECT attributes FROM partysvc.business_attributes "
```
pricem14pc, line 669: Can this explicitly select the`name` and `trading_as` text fields from the table instead of the `attributes` json fields. I think these were extracted to their own fields to save on the (admittedly minor) effort to map/extract from the equivalent json. Not a great reason here, but there's also a future pathway where we stop even holding the raw json.

Line 47 and 48 above would also need a tweak if we use the explicit text fields.

openapi.yaml
LJBabbage: get requests can't have requestBody according to my api editor

ras_party/controllers/queries.py
LJBabbage: I hadn't even noticed that, how odd, but yes they can

pricem14pc:
- reviewed code
- tested locally in docker and in dev environments(run time and build) and successfully tested multiple enrollment permutations
- checked cross-app usage of `v1/enrolments/respondent` is for a limited number of frontstage use cases only
- tested a number of preprod use cases for large enrolments
- native query correctly returning significantly less rows than the inefficient SQLAlchemy query (i.e. no longer outer join to `respondent`/`respondent_business` returning rogue rows for other respondents)
- tested the query explain plan and it no longer needs to create expensive hash tables or nested outer joins

openapi.yaml
pricem14pc: > get requests can't have requestBody according to my api editor

As discussed, that's probably best practice and technically correct (albeit not strictly enforced by the frameworks). As there are only three optional filter parameters they could be moved to the query string. Happy to raise a separate card for this.

pricem14pc (approved):
Happy with the current improvements and have discussed progressing the outstanding comments at a later date.

PR #452, opened 2025-02-06
Add logging and exception handling to delete respondents

pricem14pc:
# What and why?

Introduces exception handling, db commits, db rollbacks, logging and unit tests for handling batch respondent deletions.

The current implementation terminates mid process if a _single_ DB error is encountered when deleting records. This indiscriminately rolls back previous unrelated respondent record deletions.

As each respondent is unrelated there is no need to stop the process, only to roll back the failing respondent transaction and move onto the next respondent being deleted. We should commit successful deletion transactions before moving onto the next respondent being deleted.

# How to test?

This can be complicated to test due to the permutations of data required, and the potential exceptions that can be thrown. However, an approach is detailed below;

- create four separate accounts
- transfer a survey for account numbers 1 and 3 (i.e create `pending_surveys` for them)
- edit the `auth.user.last_login_date` year of all four respondents from `2025` to `2021` to make them subject to deletion
- run the following CURL commands to action the deletions
```
curl -u admin:secret -X DELETE http://localhost:8041/api/batch/account/users/mark-for-deletion
curl -u admin:secret -X DELETE http://localhost:8041/api/batch/account/users
curl -u admin:secret -X DELETE http://localhost:8081/party-api/v1/batch/respondents
```
- test the behaviour of the current implementation to replicate the bug whereby processing terminates at the first db error (also note the rollback of previously successfully processed deletions)
- test the behaviour of this PR fix 
- check the data to ensure `partysvc.respondents` 2 and 4 are deleted, and also their referenced child records
- check the data to ensure records for respondents 1 and 3 are NOT deleted, nor are their referenced child records
- check the party application logs to ensure all deletions, failed deletions, and counts are accurately logged

# Jira

https://jira.ons.gov.uk/browse/RAS-1460

### DIFF
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -131,12 +131,15 @@ def delete_respondents_marked_for_deletion(session):
     """
     respondents = session.query(Respondent).filter(Respondent.mark_for_deletion == True)  # noqa
     for respondent in respondents:
+        bound_logger = logger.bind(email=obfuscate_email(respondent.email_address, respondent_id=respondent.id))
+        bound_logger.info("Attempting to delete respondent records")
         try:
             session.query(Enrolment).filter(Enrolment.respondent_id == respondent.id).delete()
             session.query(BusinessRespondent).filter(BusinessRespondent.respondent_id == respondent.id).delete()
             session.query(PendingEnrolment).filter(PendingEnrolment.respondent_id == respondent.id).delete()
             session.query(Respondent).filter(Respondent.id == respondent.id).delete()
             send_account_deletion_confirmation_email(respondent.email_address, respondent.first_name)
+            bound_logger.info("Respondent records deleted successfully")
         except IntegrityError as e:
             logger.error(
                 "An data constraint violation occurred trying to delete the respondent records",
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -131,7 +131,7 @@ def delete_respondents_marked_for_deletion(session):
     """
     respondents = session.query(Respondent).filter(Respondent.mark_for_deletion == True)  # noqa
     for respondent in respondents:
-        bound_logger = logger.bind(email=obfuscate_email(respondent.email_address, respondent_id=respondent.id))
+        bound_logger = logger.bind(email=obfuscate_email(respondent.email_address), respondent_id=respondent.id)
         bound_logger.info("Attempting to delete respondent records")
         try:
             session.query(Enrolment).filter(Enrolment.respondent_id == respondent.id).delete()
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -148,7 +148,7 @@ def delete_respondents_marked_for_deletion(session):
             send_account_deletion_confirmation_email(respondent.email_address, respondent.first_name)
             bound_logger.info("Respondent records deleted successfully")
         except IntegrityError as e:
-            logger.error(
+            bound_logger.error(
                 "A data constraint violation occurred trying to delete the respondent records",
                 respondent_id=respondent.id,
                 party_uuid=respondent.party_uuid,
@@ -157,7 +157,7 @@ def delete_respondents_marked_for_deletion(session):
             session.rollback()
             failed_deletion_count += 1
         except SQLAlchemyError as e:
-            logger.error(
+            bound_logger.error(
                 "An error occurred trying to delete the respondent records",
                 respondent_id=respondent.id,
                 party_uuid=respondent.party_uuid,
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -131,7 +131,9 @@ def delete_respondents_marked_for_deletion(session):
     """
     respondents = session.query(Respondent).filter(Respondent.mark_for_deletion == True)  # noqa
     for respondent in respondents:
-        bound_logger = logger.bind(email=obfuscate_email(respondent.email_address), respondent_id=respondent.id)
+        bound_logger = logger.bind(email=obfuscate_email(respondent.email_address),
+                                   respondent_id=respondent.id,
+                                   party_uuid=respondent.party_uuid)
         bound_logger.info("Attempting to delete respondent records")
         try:
             session.query(Enrolment).filter(Enrolment.respondent_id == respondent.id).delete()
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -146,13 +146,17 @@ def delete_respondents_marked_for_deletion(session):
             bound_logger.info("Respondent records deleted successfully")
         except IntegrityError as e:
             logger.error(
-                "An data constraint violation occurred trying to delete the respondent records",
-                respondent_id=Respondent.id,
+                "A data constraint violation occurred trying to delete the respondent records",
+                respondent_id=respondent.id,
+                party_uuid=respondent.party_uuid,
                 error=str(e),
             )
         except SQLAlchemyError as e:
             logger.error(
-                "An error occurred trying to delete the respondent records", respondent_id=Respondent.id, error=str(e)
+                "An error occurred trying to delete the respondent records",
+                respondent_id=respondent.id,
+                party_uuid=respondent.party_uuid,
+                error=str(e)
             )
 
 
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -122,7 +122,6 @@ def update_respondent_mark_for_deletion(email: str, session):
         return "respondent does not exist", 404
 
 
-@with_db_session
 def delete_respondents_marked_for_deletion(session):
     """
     Deletes all the existing respondents and there associated data which are marked for deletion
@@ -142,10 +141,7 @@ def delete_respondents_marked_for_deletion(session):
         )
         bound_logger.info("Attempting to delete respondent records")
         try:
-            session.query(Enrolment).filter(Enrolment.respondent_id == respondent.id).delete()
-            session.query(BusinessRespondent).filter(BusinessRespondent.respondent_id == respondent.id).delete()
-            session.query(PendingEnrolment).filter(PendingEnrolment.respondent_id == respondent.id).delete()
-            session.query(Respondent).filter(Respondent.id == respondent.id).delete()
+            delete_respondent_records(session, respondent)
             send_account_deletion_confirmation_email(respondent.email_address, respondent.first_name)
             bound_logger.info("Respondent records deleted successfully")
         except IntegrityError as e:
@@ -162,6 +158,23 @@ def delete_respondents_marked_for_deletion(session):
                 party_uuid=respondent.party_uuid,
                 error=str(e),
             )
+
+
+@with_db_session
+def delete_respondent_records(session, respondent):
+    """
+    Deletes all records associated with a respondent.
+
+    :param session: The db session
+    :param respondent: The respondent whose records are to be deleted
+    :type respondent: Respondent
+    :raises IntegrityError: If there is a data constraint violation
+    :raises SQLAlchemyError: If there is a general SQLAlchemy error
+    """
+    session.query(Enrolment).filter(Enrolment.respondent_id == respondent.id).delete()
+    session.query(BusinessRespondent).filter(BusinessRespondent.respondent_id == respondent.id).delete()
+    session.query(PendingEnrolment).filter(PendingEnrolment.respondent_id == respondent.id).delete()
+    session.query(Respondent).filter(Respondent.id == respondent.id).delete()
 
 
 def send_account_deletion_confirmation_email(email_address: str, name: str):
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -144,7 +144,10 @@ def delete_respondents_marked_for_deletion(session):
         )
         bound_logger.info("Attempting to delete respondent records")
         try:
-            delete_respondent_records(session, respondent)
+            session.query(Enrolment).filter(Enrolment.respondent_id == respondent.id).delete()
+            session.query(BusinessRespondent).filter(BusinessRespondent.respondent_id == respondent.id).delete()
+            session.query(PendingEnrolment).filter(PendingEnrolment.respondent_id == respondent.id).delete()
+            session.query(Respondent).filter(Respondent.id == respondent.id).delete()
             session.commit()
             send_account_deletion_confirmation_email(respondent.email_address, respondent.first_name)
             bound_logger.info("Respondent records deleted successfully")
@@ -167,22 +170,6 @@ def delete_respondents_marked_for_deletion(session):
             session.rollback()
             failed_deletion_count += 1
     logger.info("Respondent record deletions complete", failed_deletion_count=failed_deletion_count)
-
-
-def delete_respondent_records(session, respondent):
-    """
-    Deletes all records associated with a respondent.
-
-    :param session: The transactional db session that can be rolled back
-    :param respondent: The respondent whose records are to be deleted
-    :type respondent: Respondent
-    :raises IntegrityError: If there is a data constraint violation
-    :raises SQLAlchemyError: If there is a general SQLAlchemy error
-    """
-    session.query(Enrolment).filter(Enrolment.respondent_id == respondent.id).delete()
-    session.query(BusinessRespondent).filter(BusinessRespondent.respondent_id == respondent.id).delete()
-    session.query(PendingEnrolment).filter(PendingEnrolment.respondent_id == respondent.id).delete()
-    session.query(Respondent).filter(Respondent.id == respondent.id).delete()
 
 
 def send_account_deletion_confirmation_email(email_address: str, name: str):
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -4,7 +4,7 @@ from uuid import UUID
 
 import structlog
 from flask import current_app, session
-from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.exc import IntegrityError, SQLAlchemyError
 from werkzeug.exceptions import BadRequest, NotFound
 
 from ras_party.controllers.account_controller import (
@@ -131,11 +131,22 @@ def delete_respondents_marked_for_deletion(session):
     """
     respondents = session.query(Respondent).filter(Respondent.mark_for_deletion == True)  # noqa
     for respondent in respondents:
-        session.query(Enrolment).filter(Enrolment.respondent_id == respondent.id).delete()
-        session.query(BusinessRespondent).filter(BusinessRespondent.respondent_id == respondent.id).delete()
-        session.query(PendingEnrolment).filter(PendingEnrolment.respondent_id == respondent.id).delete()
-        session.query(Respondent).filter(Respondent.id == respondent.id).delete()
-        send_account_deletion_confirmation_email(respondent.email_address, respondent.first_name)
+        try:
+            session.query(Enrolment).filter(Enrolment.respondent_id == respondent.id).delete()
+            session.query(BusinessRespondent).filter(BusinessRespondent.respondent_id == respondent.id).delete()
+            session.query(PendingEnrolment).filter(PendingEnrolment.respondent_id == respondent.id).delete()
+            session.query(Respondent).filter(Respondent.id == respondent.id).delete()
+            send_account_deletion_confirmation_email(respondent.email_address, respondent.first_name)
+        except IntegrityError as e:
+            logger.error(
+                "An data constraint violation occurred trying to delete the respondent records",
+                respondent_id=Respondent.id,
+                error=str(e),
+            )
+        except SQLAlchemyError as e:
+            logger.error(
+                "An error occurred trying to delete the respondent records", respondent_id=Respondent.id, error=str(e)
+            )
 
 
 def send_account_deletion_confirmation_email(email_address: str, name: str):
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -127,6 +127,10 @@ def delete_respondents_marked_for_deletion(session):
     """
     Deletes all the existing respondents and there associated data which are marked for deletion
 
+    NOTE: We don't delete pending_surveys records as these are subject to their own scheduled deletion
+    An IntegrityError exception will be logged if the respondent record cannot be deleted due
+    to existing pending_surveys records.
+
     :param session A db session
     """
     respondents = session.query(Respondent).filter(Respondent.mark_for_deletion == True)  # noqa
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -145,6 +145,7 @@ def delete_respondents_marked_for_deletion(session):
         bound_logger.info("Attempting to delete respondent records")
         try:
             delete_respondent_records(session, respondent)
+            session.commit()
             send_account_deletion_confirmation_email(respondent.email_address, respondent.first_name)
             bound_logger.info("Respondent records deleted successfully")
         except IntegrityError as e:
@@ -182,7 +183,6 @@ def delete_respondent_records(session, respondent):
     session.query(BusinessRespondent).filter(BusinessRespondent.respondent_id == respondent.id).delete()
     session.query(PendingEnrolment).filter(PendingEnrolment.respondent_id == respondent.id).delete()
     session.query(Respondent).filter(Respondent.id == respondent.id).delete()
-    session.commit()
 
 
 def send_account_deletion_confirmation_email(email_address: str, name: str):
--- a/ras_party/controllers/respondent_controller.py
+++ b/ras_party/controllers/respondent_controller.py
@@ -125,7 +125,7 @@ def update_respondent_mark_for_deletion(email: str, session):
 @with_db_session
 def delete_respondents_marked_for_deletion(session):
     """
-    Deletes all the existing respondents and there associated data which are marked for deletion
+    Deletes all the existing respondents and their associated data which are marked for deletion
 
     NOTE: We don't delete pending_surveys records as these are subject to their own scheduled deletion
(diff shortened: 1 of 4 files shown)


ras_party/controllers/respondent_controller.py
```diff
@@ -131,11 +131,22 @@ def delete_respondents_marked_for_deletion(session):
...
-        session.query(BusinessRespondent).filter(BusinessRespondent.respondent_id == respondent.id).delete()
-        session.query(PendingEnrolment).filter(PendingEnrolment.respondent_id == respondent.id).delete()
-        session.query(Respondent).filter(Respondent.id == respondent.id).delete()
-        send_account_deletion_confirmation_email(respondent.email_address, respondent.first_name)
+        try:
+            session.query(Enrolment).filter(Enrolment.respondent_id == respondent.id).delete()
+            session.query(BusinessRespondent).filter(BusinessRespondent.respondent_id == respondent.id).delete()
+            session.query(PendingEnrolment).filter(PendingEnrolment.respondent_id == respondent.id).delete()
+            session.query(Respondent).filter(Respondent.id == respondent.id).delete()
+            send_account_deletion_confirmation_email(respondent.email_address, respondent.first_name)
```
pricem14pc, line 139: Nothing handled the any exceptions thrown in the above deletions. This makes trouble shooting errors impossible, and results in the user being told their account has been deleted. Technically it can't be used, but we need sight of any issues here. Might need to alert on these log entries.

ras_party/controllers/respondent_controller.py
```diff
@@ -125,17 +125,63 @@ def update_respondent_mark_for_deletion(email: str, session):
...
+            bound_logger.info("Respondent records deleted successfully")
+        except IntegrityError as e:
+            logger.error(
+                "A data constraint violation occurred trying to delete the respondent records",
+                respondent_id=respondent.id,
+                party_uuid=respondent.party_uuid,
+                error=str(e),
+            )
+            session.rollback()
+            failed_deletion_count += 1
```
pricem14pc, line 158: If there is an exception deleting records, we `rollback()` any deleted records for that respondent, before moving onto the next respondent, maintain a sanity count of the failures.

ras_party/controllers/respondent_controller.py
```diff
@@ -125,17 +125,63 @@ def update_respondent_mark_for_deletion(email: str, session):
...
+        except SQLAlchemyError as e:
+            logger.error(
+                "An error occurred trying to delete the respondent records",
+                respondent_id=respondent.id,
+                party_uuid=respondent.party_uuid,
+                error=str(e),
+            )
+            session.rollback()
+            failed_deletion_count += 1
```
pricem14pc, line 167: As above

ras_party/controllers/respondent_controller.py
```diff
@@ -125,17 +125,63 @@ def update_respondent_mark_for_deletion(email: str, session):
...
+    :param respondent: The respondent whose records are to be deleted
+    :type respondent: Respondent
+    :raises IntegrityError: If there is a data constraint violation
+    :raises SQLAlchemyError: If there is a general SQLAlchemy error
+    """
+    session.query(Enrolment).filter(Enrolment.respondent_id == respondent.id).delete()
+    session.query(BusinessRespondent).filter(BusinessRespondent.respondent_id == respondent.id).delete()
+    session.query(PendingEnrolment).filter(PendingEnrolment.respondent_id == respondent.id).delete()
+    session.query(Respondent).filter(Respondent.id == respondent.id).delete()
+    session.commit()
```
pricem14pc, line 176: If the above deletions are successful this commit gets actioned, thus ensuring that respondent is really deleted regardless of any upcoming failures for other respondents.

ras_party/controllers/respondent_controller.py
```diff
@@ -125,17 +125,64 @@ def update_respondent_mark_for_deletion(email: str, session):
 @with_db_session
 def delete_respondents_marked_for_deletion(session):
```
pricem14pc, line 126: This controller method is invoked by a [batch HTTP endpoint](https://github.com/ONSdigital/ras-party/blob/main/ras_party/views/batch_request.py#L34-L41).

As nothing handles exceptions thrown in the above deletions, and the method doesn't throw exceptions, this makes troubleshooting errors impossible, and results in the user being told their account has been deleted but that could be rolled back when subsequent respondent deletions fail. Technically the account is unusable as it will have deleted the auth record, but we need sight of party deletions issues here. We _might_ need to alert on these log entries in the future.

test/test_respondent_controller.py
```diff
@@ -2032,3 +2035,84 @@ def _enroll_respondent(self, session):
...
+        respondent_2.first_name = "Two"
+
+        respondent_3 = Respondent()
+        respondent_3.email_address = "test3@example.com"
+        respondent_3.first_name = "Three"
+
+        respondents_to_delete = [respondent_1, respondent_2, respondent_3]
+
+        # Test currently failing, I need to work out how to throw an exception ONLY for respondent_2
+        mock_session.side_effect = IntegrityError("Integrity error", "params", "orig")
```
pricem14pc: This is correctly failing at the moment. I'm trying to work out how to throw an exception _only_ for respondent_2 to replicate the production bug.

test/test_respondent_controller.py
```diff
@@ -2032,3 +2035,84 @@ def _enroll_respondent(self, session):
...
+        # Verify
+        expected_confirmation_email_calls = [
+            call(respondent_1.email_address, respondent_1.first_name),
+            call(respondent_2.email_address, respondent_2.first_name),
+            call(respondent_3.email_address, respondent_3.first_name),
+        ]
+        mock_send_account_deletion_confirmation_email.assert_has_calls(expected_confirmation_email_calls)
+        self.assertEqual(mock_send_account_deletion_confirmation_email.call_count, 2)
+        self.assertEqual(mock_session.commit.call_count, 2)
+        self.assertEqual(mock_session.rollback.commit.call_count, 1)
```
pricem14pc: This will test the fix for the production bug. Essentially we want respondent_1 and respondent_3 to be emailed and their deletions committed, but respondent_2 should have its deletions rolled back.

ras_party/controllers/respondent_controller.py
```diff
@@ -125,17 +125,64 @@ def update_respondent_mark_for_deletion(email: str, session):
...
-        send_account_deletion_confirmation_email(respondent.email_address, respondent.first_name)
+        bound_logger = logger.bind(
+            email=obfuscate_email(respondent.email_address),
+            respondent_id=respondent.id,
+            party_uuid=respondent.party_uuid,
+        )
+        bound_logger.info("Attempting to delete respondent records")
+        try:
+            delete_respondent_records(session, respondent)
+            session.commit()
```
pricem14pc: If there are no exceptions deleting records, we `commit()` those deleted records, email the respondent, before moving onto the next respondent.

test/test_respondent_controller.py
```diff
@@ -2032,3 +2035,84 @@ def _enroll_respondent(self, session):
...
+        respondent_3 = Respondent()
+        respondent_3.email_address = "test3@example.com"
+        respondent_3.first_name = "Three"
+
+        respondents_to_delete = [respondent_1, respondent_2, respondent_3]
+
+        # Ideally this test would only throw an exception for respondent_2
+        # so respondent_1 and respondent_3 would be committed and respondent_2 would be rolled back
+        # but I can't work out how to do that yet so we're throwing exceptions for all 3
+        # this is good enough to test the production bug fix as previously it was terminating after a single exception
```
pricem14pc: I can't find a way to conditionally raise an exception in the class under test ONLY for respondent_2 so I just mock raising an exception for all three. Ideally I'd want to assert that two respondents `commit()` and one respondent `rollback()` showing that the process continues. This test does indeed show the process continues, but it does that while raising exceptions for every respondent under test (which may indeed happen so its still a valid test)

ras_party/controllers/respondent_controller.py
```diff
@@ -125,17 +125,64 @@ def update_respondent_mark_for_deletion(email: str, session):
 @with_db_session
```
LJBabbage, line 125: Your session management and interaction with the decorator is slightly odd now. You are affectively going to run commit twice. You feasibly could remove the decorator and  manage it yourself if this is the pattern you want

ras_party/controllers/respondent_controller.py
```diff
@@ -125,17 +125,64 @@ def update_respondent_mark_for_deletion(email: str, session):
...
+    Deletes all the existing respondents and their associated data which are marked for deletion
+
+    NOTE: We don't delete pending_surveys records as these are subject to their own scheduled deletion
+    An IntegrityError exception will be logged if the respondent record cannot be deleted due
+    to existing pending_surveys records.
 
     :param session A db session
     """
     respondents = session.query(Respondent).filter(Respondent.mark_for_deletion == True)  # noqa
+    logger.info("Preparing to delete respondent records", number_to_delete=respondents.count())
```
LJBabbage, line 137: Very robust logging going on here. With party the way it is I think we should look to make this a little more streamlined, the attempting bit in my opinion can be removed, the is/isn't deleted covers this. Personally I would focus just on failures and worry less about the count

ras_party/controllers/respondent_controller.py
```diff
@@ -125,17 +125,64 @@ def update_respondent_mark_for_deletion(email: str, session):
...
+        except IntegrityError as e:
+            logger.error(
+                "A data constraint violation occurred trying to delete the respondent records",
+                respondent_id=respondent.id,
+                party_uuid=respondent.party_uuid,
+                error=str(e),
+            )
+            session.rollback()
+            failed_deletion_count += 1
+        except SQLAlchemyError as e:
```
LJBabbage, line 159: Do we need both if you are including the error?

ras_party/controllers/respondent_controller.py
```diff
@@ -125,17 +125,64 @@ def update_respondent_mark_for_deletion(email: str, session):
...
+def delete_respondent_records(session, respondent):
+    """
+    Deletes all records associated with a respondent.
+
+    :param session: The transactional db session that can be rolled back
+    :param respondent: The respondent whose records are to be deleted
+    :type respondent: Respondent
+    :raises IntegrityError: If there is a data constraint violation
+    :raises SQLAlchemyError: If there is a general SQLAlchemy error
+    """
```
LJBabbage: More explicit then we normal would do and with a few assumptions. This is a standalone function and might not be rolled back from the thing calling it in the future. It probably should be private and feasibly should be rationalised with delete_respondent_by_email. I always type hint and it seems odd you list possible exceptions but not return

test/test_respondent_controller.py
```diff
@@ -2032,3 +2035,84 @@ def _enroll_respondent(self, session):
...
+        respondent_2.email_address = "test2@example.com"
+        respondent_2.first_name = "Two"
+
+        respondent_3 = Respondent()
+        respondent_3.email_address = "test3@example.com"
+        respondent_3.first_name = "Three"
+
+        respondents_to_delete = [respondent_1, respondent_2, respondent_3]
+
+        # Ideally this test would only throw an exception for respondent_2
```
LJBabbage: side_effect can use a list https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect

LJBabbage (changes requested):

ras_party/controllers/respondent_controller.py
pricem14pc: The outer method (decorated) establishes a base transaction, currently this solely controls the persistence of any transactions within that method (hence the bug in production).

However, the outer method needs (at the very least) a non-transactional session as it makes a query to get the list of respondents in the first place. Rather than handling two separate sessions, it seemed reasonable to reuse the same one. It also reasonable to control the rollback or commit of individual statements within that transaction, even though the app hander does indeed commit when control is handed back. But yes, that final commit is redundant regardless.

Agreed though, preferably not using the decorator would make more sense.

test/test_respondent_controller.py
pricem14pc: Yeah, I'd already tried this but it didn't seem to work
```
        def side_effect_function(*args, **kwargs):
            if args[0] == respondent_2:
                raise IntegrityError("Integrity error", "params", "orig")
            return MagicMock()

mock_session.side_effect = side_effect_function
```
Didn't have time to play around too much but might take another look

ras_party/controllers/respondent_controller.py
pricem14pc: This was intended to be a local private method for this purpose only. The known exceptions are handled by the calling method to control continued processing. Any other exceptions will be handled by Flask and control will stop as it currently does (inc. rollback)

I had more going on here originally, but looking at it now I might just move these 4 lines back to the calling method.

ras_party/controllers/respondent_controller.py
pricem14pc: I handled the integrity error specifically as it is the cause of the known bug. And we think we understand why its happening (business process is allowing it). I then handled any other SQLAlchemyError separately as we don't know what other things might be problematic. This was a simple way to expose that. Any other exceptions will bubble back to Flask, and stop the whole process as it currently does.

ras_party/controllers/respondent_controller.py
pricem14pc: Yeah, agreed. But this is very much to expose the extent of the issue, and any other scenarios that we might be suffering from. Currently we have zero sight of the extent of the problem. This can be changed to debug after a period of time once we're happy. We needs counts here to align with auth deletions from the previous evening. I don't want to be counting log entries manually. Due to nature of deletions, there's obviously no db records we can refer to for this.

ras_party/controllers/respondent_controller.py
pricem14pc: I meant it to be private. Is that what the a preceding underscore does? Shame Python doesn't have explicit declarations :o)

test/test_respondent_controller.py
LJBabbage: You usually interact with it more simply and use the index of the call ie..  mock_session.side_effect = [your positive return, IntegrityError(), your positive return]

pricem14pc: Thanks Lewis, I'll take another look.

pricem14pc: Finally managed to get this working, so there's now a test exercising the production bug this PR is intending to fix.

test/test_respondent_controller.py
```diff
@@ -2032,3 +2035,126 @@ def _enroll_respondent(self, session):
...
+        mock_logger.info.assert_called_with("Respondent record deletions complete", failed_deletion_count=0)
+        mock_session.rollback.assert_not_called()
+
+    @patch(
+        "ras_party.controllers.respondent_controller.send_account_deletion_confirmation_email", new_callable=MagicMock
+    )
+    @patch("ras_party.controllers.respondent_controller._delete_respondent_records")
+    @patch("ras_party.controllers.respondent_controller.session", new_callable=MagicMock)
+    @patch("ras_party.controllers.respondent_controller.logger.bind", new_callable=MagicMock)
+    def test_delete_respondents_continues_when_exception(
```
pricem14pc, line 2083: This test is specifically designed to test the production bug

SteveScorfield (approved):
Tested and works correctly. No more errors relating to rollbacks. Code looks good. Approved.

test/test_respondent_controller.py
```diff
@@ -2032,3 +2035,126 @@ def _enroll_respondent(self, session):
...
+        respondent_3 = Respondent()
+        respondent_3.email_address = "test3@example.com"
+        respondent_3.first_name = "Three"
+        respondent_3.id = 3
+        respondent_3.party_uuid = "a2a03e6f-89b3-49c2-af82-4bd411ef3307"
+
+        respondents_to_delete = [respondent_1, respondent_2, respondent_3]
+
+        mock_query = MagicMock()
+        mock_query.filter(Respondent.mark_for_deletion).return_value = respondents_to_delete
```
LJBabbage: This looks wrong to me and can be removed

test/test_respondent_controller.py
```diff
@@ -2032,3 +2035,126 @@ def _enroll_respondent(self, session):
...
+        # To iterate over "for respondent in respondents" the __iter__ method needs defining for the Mock
+        mock_query.filter.return_value.__iter__.return_value = iter(respondents_to_delete)
+
+        # Test when the second respondent of three has pending_surveys
+        def side_effect_function(respondent, session):
```
LJBabbage: Session isn't a used var so you can just placeholder it ```def side_effect_function(respondent, _)```

test/test_respondent_controller.py
```diff
@@ -2032,3 +2035,126 @@ def _enroll_respondent(self, session):
...
+            if respondent.id == 2:  # Assuming respondent_2.id is 2
```
LJBabbage: It's not an assumptions you defined it above

LJBabbage: ```mock_delete_respondent_records.side_effect = [None, IntegrityError("Integrity error", "params", "orig"), None]``` is much cleaner and works

test/test_respondent_controller.py
```diff
@@ -2032,3 +2035,126 @@ def _enroll_respondent(self, session):
...
+        self.assertEqual(mock_session.commit.call_count, 3)
+        mock_logger.info.assert_called_with("Respondent record deletions complete", failed_deletion_count=0)
+        mock_session.rollback.assert_not_called()
+
+    @patch(
+        "ras_party.controllers.respondent_controller.send_account_deletion_confirmation_email", new_callable=MagicMock
+    )
+    @patch("ras_party.controllers.respondent_controller._delete_respondent_records")
+    @patch("ras_party.controllers.respondent_controller.session", new_callable=MagicMock)
+    @patch("ras_party.controllers.respondent_controller.logger.bind", new_callable=MagicMock)
```
LJBabbage: This one doesn't need new_callable=MagicMock

test/test_respondent_controller.py
```diff
@@ -2032,3 +2035,126 @@ def _enroll_respondent(self, session):
...
+        respondent_3.first_name = "Three"
+        respondent_3.id = 3
+        respondent_3.party_uuid = "a2a03e6f-89b3-49c2-af82-4bd411ef3307"
+
+        respondents_to_delete = [respondent_1, respondent_2, respondent_3]
+
+        mock_query = MagicMock()
+        mock_query.filter(Respondent.mark_for_deletion).return_value = respondents_to_delete
+        # To iterate over "for respondent in respondents" the __iter__ method needs defining for the Mock
+        mock_query.filter.return_value.__iter__.return_value = iter(respondents_to_delete)
```
LJBabbage: It can now be mock_query.filter.return_value.__iter__.return_value = [respondent_1, respondent_2, respondent_3]

test/test_respondent_controller.py
```diff
@@ -2032,3 +2035,126 @@ def _enroll_respondent(self, session):
...
+            call("test2@example.com", "Two"),
+            call("test3@example.com", "Three"),
+        ]
+        mock_send_account_deletion_confirmation_email.assert_has_calls(expected_confirmation_email_calls)
+        self.assertEqual(mock_session.commit.call_count, 3)
+        mock_logger.info.assert_called_with("Respondent record deletions complete", failed_deletion_count=0)
+        mock_session.rollback.assert_not_called()
+
+    @patch(
+        "ras_party.controllers.respondent_controller.send_account_deletion_confirmation_email", new_callable=MagicMock
```
LJBabbage: This isn't needed either new_callable=MagicMock

LJBabbage (changes requested):

test/test_respondent_controller.py
pricem14pc: Removed, was left over from a previous approach.

test/test_respondent_controller.py
pricem14pc: While it might look cleaner to you, it's now slightly obscured what it's doing. The side effect function to me was more explicit as to the intention.

pricem14pc: > It's not an assumptions you defined it above

Left over from previous POC

LJBabbage (changes requested):
The logging does work, although as the failed_deletion_count is shared and inside an info it kind of gets hidden, however you do have the "A data constraint violation occurred trying to delete the respondent records" as an error which helps. On that subject you don't have to punch out the party_id again as it is already in the error detail.

Just logging this is fine, but the process is not so good. I lose my respondent in auth regardless, so the account becomes unusable  even though it doesn't get deleted in party. 

I also lose my shared to target, which conflicts with the expectations in the instructions, its the pending_surveys_shared_by_fkey which causes the foreign key constraint which would make sense, that I keep the person that is sharing, but the one shared to is removed. This then will of course also leave a record in pending_surveys to an email that has been deleted.

I do get one email sent which is kind of right with how it is working, rather than the expected behaviour. However the email to the one that has been shared to will then force a new user generation, which might not be quite right 

In summary logging is fine, the data that is left isn't great and we definitely need to priorities an alert if we want to merge

pricem14pc:
> I lose my respondent in auth regardless, so the account becomes unusable even though it doesn't get deleted in party.

This is intended existing functionality and is correct. The auth user account is deleted on night 1, sending a request to mark the respondent party for deletion. To all intents and purposes the user account has been correctly deleted and the user can no longer login. On night 2 an attempt is made to delete the respondent party records for the user who has had their account deleted on night 1. That may or may not be successful depending on the existence of dependent party records (e.g. `pending_survey`). In most scenarios that will correct itself in up to 72 hours until the `pending_survey` records are accepted or expired. As I have mentioned, there may be other scenarios we know nothing about yet that are causing the party deletion exceptions.

The only question I see still outstanding is whether the account deletion email should be sent on night 1 following the deletion of the auth user account, rather than on night 2 following the deletion of the respondent party records.

pricem14pc:
> I also lose my shared to target, which conflicts with the expectations in the instructions, its the pending_surveys_shared_by_fkey which causes the foreign key constraint which would make sense, that I keep the person that is sharing, but the one shared to is removed. This then will of course also leave a record in pending_surveys to an email that has been deleted.

I'm not sure what scenario you tested, but there was no instruction to share a survey with another account that you are also deleting. While this may be possible, this isn't a scenario I'm aware we need to deal with. The transfer target is intended to be an unrelated/different email address to those being deleted.

pricem14pc:
> failed_deletion_count is shared and inside an info it kind of gets hidden

Yes this is reasonable but its 100% better than we currently have, however the known Integrity Violation exception we are seeing are recoverable and not an exceptional scenario. Therefore logging it as info feels appropriate.

LJBabbage:
> > I also lose my shared to target, which conflicts with the expectations in the instructions, its the pending_surveys_shared_by_fkey which causes the foreign key constraint which would make sense, that I keep the person that is sharing, but the one shared to is removed. This then will of course also leave a record in pending_surveys to an email that has been deleted.
> 
> I'm not sure what scenario you tested, but there was no instruction to share a survey with another account that you are also deleting. While this may be possible, this isn't a scenario I'm aware we need to deal with. The transfer target is intended to be an unrelated/different email address to those being deleted.

This bit, you are making them all viable for deletion
```edit the auth.user.last_login_date year of all four respondents from 2025 to 2021 to make them subject to deletion```

pricem14pc:
> > > I also lose my shared to target, which conflicts with the expectations in the instructions, its the pending_surveys_shared_by_fkey which causes the foreign key constraint which would make sense, that I keep the person that is sharing, but the one shared to is removed. This then will of course also leave a record in pending_surveys to an email that has been deleted.
> > 
> > 
> > I'm not sure what scenario you tested, but there was no instruction to share a survey with another account that you are also deleting. While this may be possible, this isn't a scenario I'm aware we need to deal with. The transfer target is intended to be an unrelated/different email address to those being deleted.
> 
> This bit, you are making them all viable for deletion `edit the auth.user.last_login_date year of all four respondents from 2025 to 2021 to make them subject to deletion`

That's correct. They should all be subject to deletion. But there's no instruction to share a survey from one of the four users with another one of the four users. Which I'm assuming from your comment is what you did.

LJBabbage:
> > > > I also lose my shared to target, which conflicts with the expectations in the instructions, its the pending_surveys_shared_by_fkey which causes the foreign key constraint which would make sense, that I keep the person that is sharing, but the one shared to is removed. This then will of course also leave a record in pending_surveys to an email that has been deleted.
> > > 
> > > 
> > > I'm not sure what scenario you tested, but there was no instruction to share a survey with another account that you are also deleting. While this may be possible, this isn't a scenario I'm aware we need to deal with. The transfer target is intended to be an unrelated/different email address to those being deleted.
> > 
> > 
> > This bit, you are making them all viable for deletion `edit the auth.user.last_login_date year of all four respondents from 2025 to 2021 to make them subject to deletion`
> 
> That's correct. They should all be subject to deletion. But there's no instruction to share a survey from one of the four users with another one of the four users. Which I'm assuming from your comment is what you did.

Transfer should be for both new and existing accounts, therefore you need an account that isn't being deleted to test as well, ergo there should be 5 in your instructions, with 4 deleted

LJBabbage (approved):

LJBabbage (changes requested):
Just seen the pipfile.lock has been updated?

pricem14pc:
> > > > > I also lose my shared to target, which conflicts with the expectations in the instructions, its the pending_surveys_shared_by_fkey which causes the foreign key constraint which would make sense, that I keep the person that is sharing, but the one shared to is removed. This then will of course also leave a record in pending_surveys to an email that has been deleted.
> > > > 
> > > > 
> > > > I'm not sure what scenario you tested, but there was no instruction to share a survey with another account that you are also deleting. While this may be possible, this isn't a scenario I'm aware we need to deal with. The transfer target is intended to be an unrelated/different email address to those being deleted.
> > > 
> > > 
> > > This bit, you are making them all viable for deletion `edit the auth.user.last_login_date year of all four respondents from 2025 to 2021 to make them subject to deletion`
> > 
> > 
> > That's correct. They should all be subject to deletion. But there's no instruction to share a survey from one of the four users with another one of the four users. Which I'm assuming from your comment is what you did.
> 
> Transfer should be for both new and existing accounts, therefore you need an account that isn't being deleted to test as well, ergo there should be 5 in your instructions, with 4 deleted

Ok, it depends what scenarios you're testing. This PR was to fix a very specific production bug whereby processing stopped when an exception occurred. Your scenario may be one way of generating that but is not needed to test the issue. You don't need an existing account to transfer surveys to someone else. You can simply transfer to `me@here.com` and this will be good enough.

LJBabbage (approved):
Agreed updating Jinga is ok this time

PR #455, opened 2025-03-25
Remove business details for email end point

LJBabbage:
# What and why?
The email endpoint does not need to return information regarding the businesses it is associated with. The businesses association code is currently very inefficient and causes issues with performance, we need to remove it where we can

# How to test?
Check the code, ironically all the tests in frontstage and party don't include the association key in it's values, so this corrects them without doing anything. Check repos and the code to make sure that there is no use of the email endpoint business association, then run acceptance tests. I also ran all cron jobs and created my own CE and enrolled on it
# Jira

### DIFF
--- a/_infra/helm/party/Chart.yaml
+++ b/_infra/helm/party/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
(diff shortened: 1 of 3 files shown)


SteveScorfield (approved):
Fully tested and works as expected. Code changes look fine. Approved.

mfcrocker (approved):

PR #456, opened 2025-04-09
Update to use get_respondent_by_party_id

mfcrocker:
# What and why?
Continued work to remove unnecessary long calls when smaller ones will suffice

# How to test?
Test out the transfer/share survey functionality

# Jira
RAS-1291

### DIFF
--- a/_infra/helm/party/Chart.yaml
+++ b/_infra/helm/party/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.5.18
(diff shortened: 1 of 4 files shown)


LJBabbage (approved):
Probably needs a slightly more thorough testing from the business, as some of these changes are in the darker corners. The lack of coverage in batch_request.py isn't great either, which would have helped here, however it's not for this PR to fix. An eyeball and a few manually tests all seemed ok

matthew-robinson-ons (approved):
Tests passing locally though coverage is quite low in some of the areas the code was changed. Tested the share and transfer functionality and was working as expected. Code changes look good.

PR #469, opened 2025-11-06
RAS-1734 Secure Message don't request business associations

LJBabbage:
# What and why?
This PR adds a new endpoint to provide the latest business details
# How to test?
Test with https://github.com/ONSdigital/ras-secure-message/pull/450, make sure secure messaging still works
# Jira

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1066,6 +1066,7 @@ paths:
               type: array
               items:
                 type: string
+                format: uuid
         responses:
           200:
             description: A list of business details
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1051,6 +1051,30 @@ paths:
                 $ref: '#/components/schemas/BusinessResponse'
         404:
           description: The business does not exist
+    /businesses/latest-business-details:
+      Post:
+        tags:
+          - businesses
+        summary: returns a list of business details
+        description: Takes a list of party_uuids and returns the latest details of each business
+        parameters:
+          - name: party_uuids
+            in: body
+            required: True
+            description: a list of party_uuids
+            schema:
+              type: array
+              items:
+                type: string
+        responses:
+          200:
+            description: A list of business details
+            content:
+              application/json:
+                schema:
+                  $ref: '#/components/schemas/LatestBusinessDetails'
+          400:
+            description: A list of party_uuids should be supplied
   /businesses/sample/link/{sample-summary-id}:
     put:
       tags:
@@ -1653,6 +1677,22 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Association'
+    LatestBusinessDetails:
+      type: array
+      items:
+        type: object
+        properties:
+          id:
+            type: string
+            format: uuid
+            example: 91e0bb4f-ffad-404c-af1b-01c87037119c
+          sampleUnitRef:
+            type: string
+            format: string
+            example: 49900000001
+          name:
+            type: string
+            example: Office for National Statistics
     SurveyShare:
       type: object
       properties:
--- a/ras_party/controllers/business_controller.py
+++ b/ras_party/controllers/business_controller.py
@@ -2,7 +2,7 @@ import logging
 import uuid
 
 import structlog
-from flask import current_app
+from flask import current_app, session
 from werkzeug.exceptions import BadRequest, NotFound
 
 from ras_party.controllers.queries import (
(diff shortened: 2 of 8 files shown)


ras_party/controllers/business_controller.py
```diff
@@ -63,6 +64,19 @@ def get_businesses_by_ids(party_uuids, session):
...
 
 
+@with_query_only_db_session
+def get_latest_business_details(party_uuids: list, session: session) -> list:
+    business_details = []
+    businesses = query_latest_business_details(session, party_uuids)
+
+    for business in businesses:
+        business_details.append(
+            {"id": business.party_uuid, "sampleUnitRef": business.business_ref, "name": business.name}
```
LJBabbage, line 74: I've directly plugged into what is currently there, and why the key sampleUnitRef is perhaps not as expected. I might raise a card to change this everywhere, but going via 3 services led me to match at this stage

ras_party/controllers/respondent_controller.py
```diff
@@ -46,7 +46,7 @@ def get_respondent_by_ids(ids, session):
     :rtype: Respondent
     """
     respondents = query_respondent_by_party_uuids(ids, session)
-    return [respondent.to_respondent_with_associations_dict() for respondent in respondents]
+    return [respondent.to_respondent_dict() for respondent in respondents]
```
LJBabbage, line 49: I've checked all service and I'm confident no-where uses the associations element

test/test_party_controller.py
```diff
@@ -350,6 +350,22 @@ def test_get_party_by_survey_id_and_enrolment_statuses_with_invalid_enrolment(se
             "B", mock_business["id"], DEFAULT_SURVEY_UUID, ["ENABLED", "PENDING"]
         )
 
+    def test_get_latest_business_details(self):
```
LJBabbage, line 353: I've followed the current test set-up, I don't like it, but in this instance I decided not to change it

anwilkie (approved):

ras_party/controllers/respondent_controller.py
arroyoAle: This breaks the respondents tab in the reporting units section in rOps

arroyoAle (changes requested):
Looks good other than one broken page which I suspect is the _only_ usage of the respondent associations dict

arroyoAle (approved):
All works now

openapi.yaml
```diff
@@ -1051,6 +1051,30 @@ paths:
...
+        description: Takes a list of party_uuids and returns the latest details of each business
+        parameters:
+          - name: party_uuids
+            in: body
+            required: True
+            description: a list of party_uuids
+            schema:
+              type: array
+              items:
+                type: string
```
pricem14pc, line 1068:
```suggestion
                type: string
                format: uuid
```

ras_party/controllers/business_controller.py
pricem14pc: If you mean the fact that the key in our business representation is `sampleUnitRef` but the value is a `business_ref` then this is technically correct and arguably doesn't need changing.

The original solution architecture specifies the `sampleUnitRef` as a high level generic sample attribute (used across the ONS and is not specific to RASRM), which for RASRM business surveys we assign the IDBR `ru_ref` (renamed `business_ref` in RASRM).

Enterprise architecture still makes reference to our RASRM services supporting non business surveys. Therefore `sampleUnitRef` (i.e. the unit ref that came in on a sample) is ok.

ras_party/controllers/respondent_controller.py
pricem14pc: What was the resolution following this comment?

pricem14pc (approved):
Other than the resolution of the outstanding conversations

ras_party/controllers/respondent_controller.py
LJBabbage: It's been fixed, there is a card in rops

ras_party/controllers/business_controller.py
LJBabbage: Naming conventions across our services are a bit muddled with camelCase and snake. Although snake is preferred in python, JSON leans towards camel. For clarity I also find naming keys after model attributes is usually best, giving a degree of continuity as you go from controller to view then into another service. This is me saying I probably wouldn't do it like this, but have matched what was currently there as it is fairly arbitrary

ras_party/controllers/respondent_controller.py
LJBabbage: It has been fixed in rops

PR #472, opened 2025-12-12
RAS-1757 Ensure Rops free text searches (e.g. RU Search) are sanitised

LJBabbage:
# What and why?
The searches are ORM so are not subject to injection, but the code in this area is poor, and needs properly rewriting. I have affectively removed the eval statement as it's not need, but not touched anything else. The original tests didn't extend to include is_ru_ref_search so I didn't have to change any
# How to test?
Check the search still works (deploy with https://github.com/ONSdigital/response-operations-ui/pull/1044)
# Jira

### DIFF
--- a/ras_party/controllers/queries.py
+++ b/ras_party/controllers/queries.py
@@ -395,43 +395,43 @@ def search_business_with_ru_ref(search_query: str, page: int, limit: int, max_re
     bound_logger = logger.bind(search_query=search_query)
     bound_logger.info("Query looks like an ru_ref, searching only on ru_ref")
     offset = (page - 1) * limit
-    if search_query.isdigit():
-        if len(search_query) == 11:
-            bound_logger.info("Searching businesses by full ru_ref with search query")
-            result = (
-                session.query(BusinessAttributes.name, BusinessAttributes.trading_as, Business.business_ref)
-                .select_from(BusinessAttributes)
-                .join(Business)
-                .filter(Business.business_ref == search_query)
-                .distinct()
-                .all()
-            )
-            return result, len(result)
 
-        else:
-            bound_logger.info("Searching businesses by partial ru_ref with search query")
-            pages = (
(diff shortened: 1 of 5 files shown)


ras_party/views/business_view.py
```diff
@@ -93,11 +93,8 @@ def get_party_by_search():
     query = request.args.get("query", "")
     page = int(request.args.get("page", default=1))
     limit = int(request.args.get("limit", default=100))
-    is_ru_ref_search = eval(request.args.get("ru_search", default="False"))
     max_rec = int(request.args.get("max_rec", default=10000))
```
LJBabbage, line 96: 10,000?, mentioned in original PR and just ignored and merged

ras_party/controllers/queries.py
```diff
@@ -395,43 +395,43 @@ def search_business_with_ru_ref(search_query: str, page: int, limit: int, max_re
...
-            estimated_total_records = pages.scalar()
-            # we don't want to overload database with the search which retrieves more than 10000 records
-            # as it's irrelevant to show so many records as a paginated search on frontend
-            # hence this 'if' logic will avoid such searches and frontend will ask the user to refine their search
-            if pages.scalar() > max_rec:
-                return [], estimated_total_records
-            return result, estimated_total_records
+
+    if len(search_query) == 11:
+        bound_logger.info("Searching businesses by full ru_ref with search query")
```
LJBabbage, line 400: This needs a proper rewrite, I just moved the if logic up

anwilkie (approved):

ras_party/views/business_view.py
arroyoAle: It is weird but we override it anyway I guess

arroyoAle (approved):

ras_party/views/business_view.py
LJBabbage: Nope, it's not passed

PR #483, opened 2026-06-23
Upgrade to python 3.14

matthew-robinson-ons:
# What and why?
Upgrades to python 3.14, required some change to the test setup and config due to how 3.14 interacts with the db. 
There were intermittent test failures when running against the local postgres instance and different intermittent failures when running the tests against a docker network postgres instance. 

Raises the point that these are unit tests and shouldn't be interacting with the db at all, however this is through out all of the party tests. The local tests are now incredibly slow, running them against the docker network is still useable but against a local postgres instance is basically unusable as it takes > 40mins to execute.

# How to test?
Test locally and in gcp. Run the acceptance tests and local tests. 
# Jira
https://officefornationalstatistics.atlassian.net/browse/RAS-1954


test/party_client.py
```diff
@@ -42,7 +42,7 @@ def create_app():
         logger_initial_config(log_level=app.config["LOGGING_LEVEL"])
         app.config["PARTY_SCHEMA"] = party_schema.schema
         app.db = create_database(app.config["DATABASE_URI"], app.config["DATABASE_SCHEMA"])
-        app.config["EMAIL_TOKEN_EXPIRY"] = 0
+        app.config["EMAIL_TOKEN_EXPIRY"] = 36000
```
matthew-robinson-ons: This was flaky anyway as it was set to 0. In python 3.14 the db interactions are more precise and this was causing intermittent test failures depending on how fast the tests executed. Updated to match the party client config value and a more realistic token expiry value.

test/test_share_survey.py
```diff
@@ -593,6 +594,7 @@ def __init__(self):
             "survey_id": DEFAULT_SURVEY_UUID,
             "shared_by": DEFAULT_RESPONDENT_UUID,
             "batch_no": uuid.uuid1(),
+            "time_shared": datetime.now(UTC) - timedelta(days=2),
```
matthew-robinson-ons, line 597: Both `TestShareSurvey` and `TestTransferSurvey` were failing on python 3.14 as the mock was being setup with `time_shared = now()` and then immediately checking if `time_shared < now` This was a bit flaky and sometimes leads to the if statement returning false in the tests. Explicitly setting time_shared in the past to prevent this.

test/test_transfer_survey.py
```diff
@@ -601,6 +602,7 @@ def __init__(self):
             "shared_by": DEFAULT_RESPONDENT_UUID,
             "batch_no": uuid.uuid1(),
             "is_transfer": True,
+            "time_shared": datetime.now(UTC) - timedelta(days=2),
```
matthew-robinson-ons, line 605: As above

config.py
```diff
@@ -1,6 +1,5 @@
 # flake8: noqa
 import os
-from distutils.util import strtobool
```
matthew-robinson-ons, line 3: Left over deprecated import, we're not using this anymore.

PR #488, opened 2026-08-06
Share postgres image version between local dev and CI

prem-buczkowski-ons:
# What and why?
Cloud Build starts a postgres container for the test suite, and needs to know which version. Rather than pinning it in two places, the version lives in _infra/postgres-image and is fed to docker-compose via --env-file, so local dev and CI stay in step. Bumps `postgres` from the EOL 9.6 to 14, matching `ras-rm-docker` and the prod.

Compose now requires that flag, so use `make start-db` rather than a bare `docker compose up`.

If you would rather take a different approach, let me know.

# How to test?
Run `make start-db` and see the version started in `docker-compose logs`.

# Jira
* https://officefornationalstatistics.atlassian.net/browse/RAS-1930
* https://officefornationalstatistics.atlassian.net/wiki/spaces/SDC/pages/303956413/Spinnaker+Removal

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,11 @@
-.PHONY: build test start
+.PHONY: build test start start-db
 
 build:
 	pipenv install --dev
+
+# The postgres image version is read from _infra/postgres-image, which CI uses too.
+start-db:
+	docker compose --env-file _infra/postgres-image up -d
 
 lint:
 	pipenv run isort .
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,8 +35,7 @@ jobs:
           python-version: '3.12'
       - name: Get Dependencies
         run: |
-          docker compose --verbose build
-          docker compose up -d
+          make start-db
           pip install pipenv
           pipenv install --dev
       - name: Run tests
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -1,8 +1,12 @@
-.PHONY: build test start
+.PHONY: build test start start-db
 
 build:
 	pipenv install --dev
 
+# The postgres image version is read from _infra/postgres-image, which CI uses too.
+start-db:
+	docker compose --env-file _infra/postgres-image up -d
```
warrenbaileyons: I can't run this make command as we don't have docker on our macbooks, we have to use podman and the make file doesn't respective aliases.

It might be worth us having a `start-local-db` target that uses `podman compose  --env-file _infra/postgres-image up -d`

warrenbaileyons (commented):
LGTM, just need to add a new make command for podman macbooks

Makefile
prem-buczkowski-ons: In SRM, we set `DOCKER ?= $(shell if [ "$$(uname -m)" = "arm64" ]; then echo podman; else echo docker; fi)` and use that to run targets. Would that work?

warrenbaileyons: Perfect lets do that 👍

prem-buczkowski-ons: Done.

warrenbaileyons (approved):

arroyoAle (approved):

